PR #5, opened 2022-11-04
new simplified project structure

RadovanTomik:


pom.xml
```diff
@@ -11,11 +11,7 @@
   <artifactId>negotiator</artifactId>
   <version>3.0.1</version>
   <modelVersion>4.0.0</modelVersion>
-  <packaging>pom</packaging>
-
-  <modules>
-    <module>negotiator-application</module>
-  </modules>
+  <packaging>war</packaging>
```
svituz, line 14: it should be used jar as packaging, or we'd need to change the Docker ENTRYPOINT that is `ENTRYPOINT ["java","-jar", "-Dspring.profiles.active=docker", "/negotiator-application-3.0.1.jar"]`

RadovanTomik: Should be fixed, please try

PR #28, opened 2022-11-14
Legacy code refactoring

svituz:
Some fixes and refactoring.

In particular:
- the default redirect uri has been changed to /request/{requestID}
- in NegotiationController the getCreator and getCreatorId have been merged into a unique method since the getCreator is not used. The method has been changed to be not static since it was generating errors when using the api (strangely, not in the unit test)
- The RequestV2Controller has been changed to QueryV2Controller, since in that version used that name
-


src/main/resources/application.yaml
```diff
@@ -15,6 +15,7 @@ spring:
         jwt:
           issuer-uri:
 negotiator:
+  frontendUrl: "http://localhost"
   redirectPath: "/gui/negotiation"
```
RadovanTomik, line 19: Could you remove the /gui?

RadovanTomik (approved):

PR #40, opened 2023-04-21
Refactoring to use DTOs as interface

svituz:
PR that refactors interface between Controllers and Services.
Now the services accept and return DTOS.
Other changes:
 - Added interfaces for Services
 - Added unit test structure with very few unit test
 - Split tests between unit and integration and moved all the old test into integration
 - Created in the configuration some classes to map between data models. Before that, the mappers were inside the Services classes. Now they are all gathered in a single package


RadovanTomik:
Please autowire interfaces not their implementations. https://stackoverflow.com/questions/12899372/spring-why-do-we-autowire-the-interface-and-not-the-implemented-class

RadovanTomik (changes requested):
Please add java doc for the interfaces and implement/fix TO DOs

RadovanTomik (approved):

PR #55, opened 2023-06-27
fix: Preserves h2 connection to close and lost in memory table data

asulis:
This PR preserves the H2 in memory database to lose "PERSON" table and cause the crash of the application. Usually, it happens when the last opened DB connection is closed: when it happens, the H2 drops the table because it has not persistence. The added parameter allows to keep the connection alive, until the Virtual Machine is up.


asulis:
Dropped branch

PR #65, opened 2023-07-21
Refactor managment of Negotiation Lifecycle

RadovanTomik:
### Description:

- Refactoring of the lifecycle management. The current lifecycle state is no longer saved in a spring state machine but rather as an attribute in the Negotiation entity and the springs state machine is only being used as a workflow configuration that also triggers the updates.
- Unit tests for AccessCriteria mapping

### Checklist:

- [x] I have performed a self review of my code
- [x] My code follows [Google Java Code style](https://google.github.io/styleguide/javaguide.html)
- [x] I have added unit tests and the code coverage has not decreased
- [x] I have updated the documentation in all relevant places


RadovanTomik (commented):
slef review done

konradlang (approved):
Approved after internal discussion.

PR #231, opened 2024-03-18
Refactor/use internal resource

asulis:
### Description:

### Checklist:

- [x] I have performed a self review of my code
- [x] My code follows [Google Java Code style](https://google.github.io/styleguide/javaguide.html)
- [x] I have added unit tests and the code coverage has not decreased
- [x] I have updated the documentation in all relevant places


src/main/java/eu/bbmri_eric/negotiator/database/model/Negotiation.java
```diff
@@ -86,7 +84,7 @@ public class Negotiation extends AuditEntity {
   @Exclude
   private Set<PersonNegotiationRole> persons = new HashSet<>();
 
-  @OneToMany(mappedBy = "negotiation", cascade = CascadeType.MERGE, fetch = FetchType.LAZY)
+  @OneToMany(mappedBy = "negotiation", cascade = CascadeType.MERGE, fetch = FetchType.EAGER)
```
RadovanTomik, line 87: not sure if eager is needed here.

src/main/java/eu/bbmri_eric/negotiator/database/model/Negotiation.java
```diff
@@ -140,15 +127,39 @@ public void setCurrentState(NegotiationState negotiationState) {
...
-      this.negotiationResourceLifecycleRecords.add(record);
+    NegotiationResourceLifecycleRecord record =
+        NegotiationResourceLifecycleRecord.builder()
+            .changedTo(state)
+            .resource(lookupResource(getResources(), resourceId))
+            .build();
+    this.negotiationResourceLifecycleRecords.add(record);
+  }
+
+  public Map<String, NegotiationResourceState> getCurrentStatePerResource() {
```
RadovanTomik, line 138: please refactor this method as it is quite long. Divide it into multiple private methods

src/main/java/eu/bbmri_eric/negotiator/database/model/Request.java
```diff
@@ -52,7 +52,7 @@ public class Request {
   @Column(name = "id")
   private String id;
 
-  @ManyToMany
+  @ManyToMany(fetch = FetchType.EAGER)
```
RadovanTomik, line 55: this will produce a large overhead. Can this be lazy?

RadovanTomik (changes requested):
Please see comments and add a db migration script.

src/main/java/eu/bbmri_eric/negotiator/database/model/Request.java
asulis: This is linked to the similar comment above and this change is due to the fact then, when the getCurrentState per resource is called by the resource lifecycle service, the negotiation object is not "complete", because it does not contain the child reference to the request (and then to the resource). So the getResources() method will throw a LAZY exception. That was one of the reasons why I had thought about implementing and calling the repository for the NegotiationResourceLifecycleRecord directly from the recource lifecucle service, moving out from the Negotiation class this logic, since now the state is not get anymore from a Negotiation's internal key-value table. Fore sure we can inspect (if you have suggestion) a way to avoid both the EAGERS and the repository implementation

src/main/java/eu/bbmri_eric/negotiator/database/model/Negotiation.java
asulis: Fixed in fae91486edd5eb4136965a0060278f579255ae20

asulis: Migration script added in 97adafaf3cc969b82f50fa25f60fdbc2431e8e0

src/main/java/eu/bbmri_eric/negotiator/database/model/Negotiation.java
```diff
@@ -168,6 +185,7 @@ public int hashCode() {
     return Objects.hash(getId());
   }
 
+  @Transactional
```
RadovanTomik, line 188: This annotations should not be present in an entity

src/main/java/eu/bbmri_eric/negotiator/database/model/Negotiation.java
RadovanTomik: should be removed

src/main/java/eu/bbmri_eric/negotiator/database/model/Request.java
RadovanTomik: please remove it a use something like a named entity graph, or a custom method in the repository. We cannot have Eager annotations on every property.

src/main/java/eu/bbmri_eric/negotiator/database/model/Negotiation.java
```diff
@@ -140,15 +127,45 @@ public void setCurrentState(NegotiationState negotiationState) {
   }
 
   public void setStateForResource(String resourceId, NegotiationResourceState state) {
```
RadovanTomik, line 129: please add javadoc for public methods

RadovanTomik (changes requested):

RadovanTomik (changes requested):
please see comments

src/main/java/eu/bbmri_eric/negotiator/database/model/Request.java
asulis: Fixed in 18f3761c23e6d941a2cd8c7053b98734ce3cd6a8

src/main/java/eu/bbmri_eric/negotiator/database/model/Negotiation.java
asulis: Fixed in 18f3761c23e6d941a2cd8c7053b98734ce3cd6a8

src/main/java/eu/bbmri_eric/negotiator/database/model/Negotiation.java
asulis: Fixed in ffd04f6a2618411ceddbaaf038bd371a7633ad6f

src/main/java/eu/bbmri_eric/negotiator/database/model/Negotiation.java
asulis: Fixed in ffd04f6a2618411ceddbaaf038bd371a7633ad6f

src/main/resources/db/migration/postgresql/V6.0__drop_resource_state_per_negotiation_table.sql
```diff
@@ -0,0 +1 @@
+drop table if exists public.resource_state_per_negotiation;
```
RadovanTomik, line 1: please add the new constraint you have created for Negotiation.sourceID

RadovanTomik (changes requested):
please see comments

src/main/resources/db/migration/postgresql/V6.0__drop_resource_state_per_negotiation_table.sql
asulis: Fixed in 41ce33a45006cdba6c6bd86db29045475b6c2e00

RadovanTomik (approved):

RadovanTomik (changes requested):
please rebase onto master and adjust to new changes

RadovanTomik (approved):

PR #354, opened 2024-06-11
refactor: remove unused person negotiation role

svituz:
## Negotiator pull request:

This PR removes unnecessary entities PersonNegotiationRole and Role.

### Description:

The PersonNegotiationRole was created to record the roles associated with a negotiation but, as a matter of fact, it was used to record only its creator, which is also registered in the createdBy attribute, thus resulting in information duplication. After removing PersonNegotiatioRole, also the Role entity was unnecessary since the role is always obtained from the token.

### Checklist:

- [X] I have performed a self review of my code
- [X] My code follows [Google Java Code style](https://google.github.io/styleguide/javaguide.html)
- [X] I have made my code as simple as possible
- [X] I have added unit tests and the code coverage has not decreased
- [X] I have updated the documentation in all relevant places


RadovanTomik (changes requested):
Please just rename migration script to v9

RadovanTomik (approved):

PR #385, opened 2024-07-24
feat: add mechanism for requiring additional information in the state machine

RadovanTomik:
### Description:

This PR adds new functionality for attaching an information requirement to Resource states that need to be fulfilled for the State machine to proceed.

### Checklist:


src/main/java/eu/bbmri_eric/negotiator/exceptions/NegotiatorExceptionHandler.java
```diff
@@ -275,4 +279,22 @@ public final ResponseEntity<HttpErrorResponseModel> handleDataIntegrityViolation
             .build();
     return new ResponseEntity<>(errorResponse, HttpStatus.BAD_REQUEST);
   }
+
+  @ExceptionHandler(NullPointerException.class)
```
svituz, line 283: I think this is too risky. NullPointerException may be a programming error and you usually don't have control over it. You would end up here each time an uncontrolled NullPointerException is thrown. If you have some point where you know it may occur then I suggest you catch it and throw another exception that you can handle here

src/main/java/eu/bbmri_eric/negotiator/exceptions/NegotiatorExceptionHandler.java
```diff
@@ -275,4 +279,22 @@ public final ResponseEntity<HttpErrorResponseModel> handleDataIntegrityViolation
...
+  @ResponseStatus(HttpStatus.BAD_REQUEST)
+  public final ErrorResponse handleNullPointerException(NullPointerException ex) {
+    ProblemDetail detail = ProblemDetail.forStatus(HttpStatus.BAD_REQUEST);
+    detail.setTitle("Encountered a null pointer exception");
```
svituz, line 287: I don't think the client wants to know about this

svituz (approved):

Issue #361, opened 2024-06-20
[FEATURE] UI configuration


RadovanTomik:
**Brief description.**
Add a REST API endpoint for fetching and updating global UI configuration. For a list of all relevant values contact @tmilost. The update operations should be alowed only for admins. Example response:
 ```
{
"primary-color": "red",
"auth_link": "https://wahever.com"
"logo_url": "https://wahever.com/icon.svg"
}

```
**Desired functionality acceptance criteria**

- [ ] GET `/ui-config`
- [ ] PATCH `/ui-config {"key": "new_value"}`

tmilost:
current values: 
`activeThemeFile: "bbmri",
  activeLogosFiles: "bbmri",
  isFooterStatusPageVisible: true,
  footerStatusPageLink: "https://status.bbmri-eric.eu/",
  footerStatusPageText: "BBMRI-ERIC Status Page",
  isFooterFollowUsVisible: true,
  footerFollowUsLinkedin: "https://www.linkedin.com/company/bbmri-eric",
  footerFollowUsX: "https://twitter.com/BBMRIERIC",
  footerFollowUsPodcast: "https://www.bbmri-eric.eu/bbmri-eric/bbmri-eric-podcast/",
  isFooterWorkProgrammeVisible: true,
  footerWorkProgrammeLink: "https://www.bbmri-eric.eu/wp-content/uploads/BBMRI-ERIC_work-program_2022-2024_DIGITAL.pdf",
  isFooterPrivacyPolicyVisible: true,
  footerPrivacyPolicy: "https://www.bbmri-eric.eu/wp-content/uploads/AoM_10_8_Access-Policy_FINAL_EU.pdf",
  isFooterHelpLinkVisible: true,
  footerHelpLink: "mailto:negotiator@helpdesk.bbmri-eric.eu",
  isFooterCopyRightVisible: true,
  footerCopyRight: "© 2024 BBMRI-ERIC",
  isFooterLeftSideIconVisible: true,
  footerLeftSideIcon: "@/assets/images/bbmri/home-bbmri-png",
  footerLeftSideIconLink: "https://www.bbmri-eric.eu/",
  isFooterNewsletterVisible: true,
  footerNewsletter: "https://www.bbmri-eric.eu/news-event/"`

PR #469, opened 2024-10-09
feat: adds API to get settings for ui

svituz:
### Description:

PR related to issue #361. It adds the /ui-config endpoint that returns a json with the configurations for a UI.
The settings parameters are stored in the DB in a specific table

### Checklist:


RadovanTomik (changes requested):
please go through the PR checklist and tick all the boxes

src/main/java/eu/bbmri_eric/negotiator/settings/UIConfigServiceImpl.java
```diff
@@ -0,0 +1,84 @@
...
+              settings
+                  .computeIfAbsent(uiParameter.getCategory(), k -> new HashMap<>())
+                  .put(uiParameter.getName(), Boolean.valueOf(uiParameter.getValue()));
+              break;
+            default:
+              settings
+                  .computeIfAbsent(uiParameter.getCategory(), k -> new HashMap<>())
+                  .put(uiParameter.getName(), uiParameter.getValue());
+              break;
+          }
```
RadovanTomik, line 45: I think this could be handled differently. Generally switch/if statements can get very long. Can't we put this logic into a private class in the Entity maybe

src/main/java/eu/bbmri_eric/negotiator/settings/UIParameterValidator.java
```diff
@@ -0,0 +1,30 @@
...
+        try {
+          Integer.parseInt(parameter.getValue());
+          return true;
+        } catch (NumberFormatException e) {
+          return false;
+        }
+      case STRING:
+        return true;
+      default:
+        return false;
```
RadovanTomik, line 27: isn't this wrong separation of concerns? If you add a new type then you have to change code in 3 places

src/main/resources/db/dev/migration/R__Initial_data.sql
```diff
@@ -61,3 +61,27 @@ insert into public.access_form_section_element_link (id, access_form_section_lin
...
+  (13, 'footer', 'footerPrivacyPolicy', 'STRING', 'https://www.bbmri-eric.eu/wp-content/uploads/AoM_10_8_Access-Policy_FINAL_EU.pdf'),
+  (14, 'footer', 'isFooterHelpLinkVisible', 'BOOL', true),
+  (15, 'footer', 'footerHelpLink', 'STRING', 'mailto:negotiator@helpdesk.bbmri-eric.eu'),
+  (16, 'footer', 'isFooterCopyRightVisible', 'BOOL', true),
+  (17, 'footer', 'footerCopyRight', 'STRING', '© 2024 BBMRI-ERIC'),
+  (18, 'footer', 'isFooterLeftSideIconVisible', 'BOOL', true),
+  (19, 'footer', 'footerLeftSideIcon',  'STRING', '@/assets/images/bbmri/home-bbmri-png'),
+  (20, 'footer', 'footerLeftSideIconLink', 'STRING', 'https://www.bbmri-eric.eu/'),
+  (21, 'footer', 'isFooterNewsletterVisible', 'BOOL', true),
+  (22, 'footer', 'footerNewsletter', 'STRING', 'https://www.bbmri-eric.eu/news-event/');
```
RadovanTomik, line 87: Shouldn't this be put into a general repeatable migration not tied to a profile? The Frontend won't work without these parameters

RadovanTomik, line 87: Shouldn't this be put into a general repeatable migration not tied to a profile? The Frontend won't work without these parameters

RadovanTomik (changes requested):

src/main/resources/db/dev/migration/R__Initial_data.sql
svituz: The configuration is specific to one UI and I think it's up to the maintainer of that one deployment to add the correct parameters for the UI. Each UI may potentially have very different parameters. This is just an example of a configuration for development that "accidentally" fits the BBMRI UI

svituz: duplicated :)

src/main/java/eu/bbmri_eric/negotiator/settings/UIParameter.java
```diff
@@ -0,0 +1,33 @@
...
+import lombok.Setter;
+
+@Getter
+@Setter
+@AllArgsConstructor
+@NoArgsConstructor
+@Entity
+@Table(name = "ui_parameter")
+@ValidUIParameter
+public class UIParameter {
```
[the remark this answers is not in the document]
svituz: added

src/main/java/eu/bbmri_eric/negotiator/settings/UIConfigServiceImpl.java
svituz: I have added a getTypedValue to the UIParameter entity. It should shift the responsibility away from the service to the entity, which is the right responsible

src/main/java/eu/bbmri_eric/negotiator/settings/UIParameterValidator.java
svituz: The only concern of this class is to check whether the parameter's value is compatible with the declared type. This is the way I found to validate an Entity instance at the class level (which is the case since the validation of the value depends on the type). 
I removed the problem of changing the service when there is a new type, but this class is by definition coupled with the entity (it's its validator), so I would keep it as it is.

src/main/java/eu/bbmri_eric/negotiator/settings/UIParameter.java
```diff
@@ -0,0 +1,33 @@
...
+  @Id Long id;
```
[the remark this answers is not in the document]
svituz: Wouldn'it be misleading?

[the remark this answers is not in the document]
RadovanTomik: I think the fact that we discard the id if it is set is misleading :D

RadovanTomik (approved):

PR #503, opened 2024-11-13
Refactor/#493 negotiation filters

svituz:
### Description:

This PR refactors the endpoints to get the list of negotiations (`/negotiations`, `/users/{id}/negotiations` and `/network/{id}/negotiations`). 

Some behaviors have changed:
 * now the endpoints don't return 401 Bad Request when an unknown filter is used in the query parameters
 * filter by role doesn't accept the `role=ROLE_ADMIN` value and returns a BadRequest now. This is a breaking change for the frontend because the route `/negotiations?role=ROLE_ADMIN` will result in an error. 
   **NB:** before this modification, the filter did nothing since the ROLE_ADMIN value wasn't already in the possible values of the `role` filter. Simply, it returned all the negotiations (correctly since the ROLE_ADMIN is authorized for all of them) because the filter was ignored. To be more precise, any value would have had the same result.
* also the `/network/{id}/negotiations` now supports the filters and the filter by network has been changed to use the NegotiationSpecification 

### Checklist:


RadovanTomik:
If I try to filter by status I get:
```json
{
  "type": "about:blank",
  "title": "Wrong request parameters",
  "status": 400,
  "detail": "{status=Failed to convert property value of type 'java.lang.String' to required type 'java.util.List' for property 'status'; Failed to convert from type [java.lang.String] to type [eu.bbmri_eric.negotiator.negotiation.state_machine.negotiation.NegotiationState] for value [[\"APPROVED\"]]}",
  "instance": "/api/v3/negotiations"
}
```

RadovanTomik (changes requested):
please see comment

RadovanTomik (approved):

PR #504, opened 2024-11-17
feat: enable editing of Negotiations

RadovanTomik:
### Description:

This PR enables the User to edit a Negotiation after it has been submitted. To test it create a Negotiation and click on edit.

### Checklist:


src/main/java/eu/bbmri_eric/negotiator/common/exceptions/NegotiatorExceptionHandler.java
```diff
@@ -48,6 +54,35 @@ public final ResponseEntity<HttpErrorResponseModel> handleJwtDecoderError(
     return new ResponseEntity<>(errorResponse, HttpStatus.INTERNAL_SERVER_ERROR);
   }
 
+  @ExceptionHandler(MaxUploadSizeExceededException.class)
+  @ApiResponse(
+      responseCode = "413",
+      description = "Payload Too Large - Max file size exceeded",
```
konradlang, line 60: Could we add the max allowed file size here, e.g. "Max file size of XXMB exceeded".

konradlang (changes requested):

src/main/java/eu/bbmri_eric/negotiator/common/exceptions/NegotiatorExceptionHandler.java
RadovanTomik: it is sent in the actual error message, this is just an example that is shown in swagger.
Screenshot 2024-11-21 at 16 12 11

konradlang (approved):

PR #521, opened 2024-12-02
feat: add message to status change for negotiations

svituz:
### Description:

This PR adds the possibility of sending a message when a status change for the negotiation is sent.
When the message is sent, it is added as a Public Post to the Negotiation.
The message is also added as part of the notification email.
Also, the list of possible events for negotiation is changed, since now it returns all the fields of the events 

### Checklist:

### DIFF
--- a/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationController.java
+++ b/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationController.java
@@ -230,6 +230,7 @@ public class NegotiationController {
                     event.getLabel(),
                     event.getDescription(),
                     event.isMessageRequired()))
+        .sorted((e1, e2) -> e1.getValue().compareTo(e2.getValue()))
         .collect(Collectors.toList());
   }
 
--- a/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationController.java
+++ b/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationController.java
@@ -6,6 +6,7 @@ import eu.bbmri_eric.negotiator.governance.resource.ResourceWithStatusAssembler;
 import eu.bbmri_eric.negotiator.governance.resource.dto.ResourceWithStatusDTO;
 import eu.bbmri_eric.negotiator.negotiation.dto.NegotiationCreateDTO;
 import eu.bbmri_eric.negotiator.negotiation.dto.NegotiationDTO;
+import eu.bbmri_eric.negotiator.negotiation.dto.NegotiationEventMetadataDTO;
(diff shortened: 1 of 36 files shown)


src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationController.java
```diff
@@ -162,17 +164,28 @@ public EntityModel<NegotiationDTO> retrieve(@Valid @PathVariable String id) {
...
+      @RequestBody(required = false) @Nullable
+          NegotiationUpdateLifecycleDTO negotiationUpdateLifecycleDTO) {
     if (!AuthenticatedUserContext.isCurrentlyAuthenticatedUserAdmin()
         && !isCreator(negotiationService.findById(id, false))) {
       throw new ResponseStatusException(HttpStatus.FORBIDDEN);
     }
     // Process the request
-    negotiationLifecycleService.sendEvent(id, event);
+    String message = null;
+    if (negotiationUpdateLifecycleDTO != null
```
RadovanTomik, line 184: Objects.isNull()

src/main/java/eu/bbmri_eric/negotiator/negotiation/state_machine/negotiation/NegotiationLifecycleService.java
```diff
@@ -25,6 +25,7 @@ public interface NegotiationLifecycleService {
    *
    * @return the new status of the Negotiation
    */
-  NegotiationState sendEvent(String negotiationId, NegotiationEvent negotiationEvent)
```
RadovanTomik, line 28: leave the orgiginal signature, just add a new method with update signature. Then the parameter will indeed be optional

RadovanTomik (changes requested):
please see comments plus I'm still not convinced we need the required parameter. It could just awayls be optional and the code would be much simpler. PLus the backend must send an email where this comment is shown. PLease see the original issue

src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationController.java
svituz: The documentation says:
This method exists to be used as a [Predicate](https://docs.oracle.com/javase/8/docs/api/java/util/function/Predicate.html), filter(Objects::isNull)

svituz:
> please see comments plus I'm still not convinced we need the required parameter. It could just awayls be optional and the code would be much simpler. PLus the backend must send an email where this comment is shown. PLease see the original issue

My understanding was that, when declining a negotiation, the administrator must specify the reason for rejection. So I came up with the messageRequired flag.

RadovanTomik (approved):

PR #562, opened 2025-01-08
fix: pie chart background color generation

tmilost:
### Description:

fix: pie chart background color generation


frontend/vite.config.js
```diff
@@ -10,14 +10,16 @@ import git from 'git-rev-sync'
 const PROXY_TARGET = "http://localhost:8081"
 
 try {
+    // eslint-disable-next-line
```
RadovanTomik, line 13: why?

RadovanTomik (approved):

PR #667, opened 2025-02-27
feat: add possibility to save the negotiation as draft before submitting it

svituz:
### Description:

This PR gives the researcher the possibility of saving a DRAFT of the negotiation.
The functionality adds the following changes:

Backend:
  - Adds the state `DRAFT` for Negotiation
  - Adds the event `SUBMIT` to change the status from `DRAFT` to `SUBMITTED`
  - Adds the `draft` flag in the `NegotiationCreateDTO`. If `true` the Negotiation is saved with `DRAFT` state

Frontend:
  - Adds the "Save Draft" button to the Negotiation Form, just in case it is a brand new request or the negotiation was already in DRAFT state (you cannot draft an already SUBMITTED negotiation)

### Checklist:


backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationSpecification.java
```diff
@@ -65,7 +68,7 @@ private static Specification<Negotiation> initOrAnd(
    * @param states a List of NegotiationState to use as filter
    * @return a Specification to add as part of a query to filter Negotiations
    */
-  public static Specification<Negotiation> hasState(List<NegotiationState> states) {
+  public static Specification<Negotiation> hasState(List<NegotiationState> states, boolean not) {
```
RadovanTomik, line 71: what is not? :D

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/state_machine/negotiation/NegotiationState.java
```diff
@@ -4,6 +4,7 @@
 
 @Getter
 public enum NegotiationState {
+  DRAFT("Draft", "The negotiation request has been started but non completed yet"),
```
RadovanTomik, line 7: we should fix the description

backend/src/test/java/eu/bbmri_eric/negotiator/integration/service/NegotiationLifecycleServiceImplTest.java
```diff
@@ -169,18 +169,20 @@ void sendEvent_abandonNegotiation_to_inProcess_Negotiation() throws IOException
             negotiationDTO.getId(), NegotiationEvent.ABANDON, "Not acceptable"));
   }
 
-  private NegotiationDTO saveNegotiation() throws IOException {
-    NegotiationCreateDTO negotiationCreateDTO = TestUtils.createNegotiation("request-2");
+  private NegotiationDTO saveNegotiation(boolean draft) throws IOException {
```
RadovanTomik, line 172: I would add it as an optional parameter, usually you wont save it as draft

RadovanTomik (changes requested):

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationSpecification.java
svituz: It is a flag used to negate the condition. I added it so I could do "hasState not DRAFT" instead of "hasState in [all other states]"

RadovanTomik: ah, then please update the javadoc so it is clear what it does or come up with a different solution

backend/src/test/java/eu/bbmri_eric/negotiator/integration/service/NegotiationLifecycleServiceImplTest.java
svituz: you mean by adding

```java
  private NegotiationDTO saveNegotiation() throws IOException {
    return saveNegotiation(false);
  }
```
?

RadovanTomik: something like this yes

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/state_machine/negotiation/NegotiationState.java
RadovanTomik: Saved as a draft, open for editing and waiting for submission?

RadovanTomik: can you please fix this? at the bare minimum there is a typo

RadovanTomik:
I cant see the save as draft button when filling out the access form

RadovanTomik:
plus lets add funtionality that will automatically save the negotiation as draft when user starts filling out the access form. Also som periodic saves and a notification that the draft was saved.

RadovanTomik (changes requested):
please see comments

svituz:
I added the automatic saving of a Negotiation. This new feature required some changes in the implementation
Indeed, the first time a draft is saved the negotiation is created and so the page must be redirected to the /edit/requests/<negotiation_id> page. 
The Save draft button is enabled only when the form has some changes and the update of the negotiation on "Next" is called only when there are some changes.
Finally a notification is sent when a negotiation is saved as draf

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/Negotiation.java
```diff
@@ -90,14 +90,13 @@ public class Negotiation extends AuditEntity {
...
-  private boolean publicPostsEnabled = true;
+  private boolean publicPostsEnabled;
 
   private boolean privatePostsEnabled = false;
 
-  @Builder.Default
   @Setter(AccessLevel.NONE)
   @Enumerated(EnumType.STRING)
-  private NegotiationState currentState = NegotiationState.SUBMITTED;
+  private NegotiationState currentState;
```
RadovanTomik, line 99: Doesnt it make sense to have a default value?

RadovanTomik (approved):
Looks good 👍

tmilost (approved):

PR #693, opened 2025-03-26
Fix file upload type validation

RadovanTomik:
### Description:

This PR adds File type validation on the API level. To test it spin up the backend and try to upload as many different attachment types as possible. http://localhost:8081/api/swagger-ui/index.html#/Attachments/create_1

### Checklist:

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -385,5 +385,10 @@
             <artifactId>commons-csv</artifactId>
             <version>1.13.0</version>
         </dependency>
+        <dependency>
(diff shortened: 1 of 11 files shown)


konradlang:
Can the file types be added to the ones accepted by the frontend also.
This should be exposed somewhere to the frontend so the available file types are also in sync with the ones accepted by the API.

tmilost (approved):

konradlang:
Now the information states the supported files types, but on windows it still is possible to upload a shell-script, without an error:
Capture-upload-shell-script

PR #694, opened 2025-03-27
feat: add customizability for email templates

jungwire:
### Description:

Adds endpoint that allow to change the email templates

### Checklist:


backend/src/main/java/eu/bbmri_eric/negotiator/TemplateInitiationRunner.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private String defaultTemplatePath = "classpath:/templates/";
+
+  @Autowired
+  public TemplateInitiationRunner(UserNotificationService userNotificationService) {
+    this.userNotificationService = userNotificationService;
+  }
+
+  @Override
+  public void run(String... args) throws Exception {
```
RadovanTomik, line 34: this code is way to complicated, we should divide it into multiple functions

backend/src/main/java/eu/bbmri_eric/negotiator/notification/NotificationController.java
```diff
@@ -28,4 +26,38 @@ public List<NotificationDTO> getNotificationsForUser() {
     return userNotificationService.getNotificationsForUser(
         AuthenticatedUserContext.getCurrentlyAuthenticatedUserInternalId());
   }
```
RadovanTomik, line 28: Not sure the notification Controller is the right place for this. Shouldn't it be separate?

backend/src/main/java/eu/bbmri_eric/negotiator/notification/NotificationController.java
```diff
@@ -28,4 +26,38 @@ public List<NotificationDTO> getNotificationsForUser() {
...
+      @PathVariable String templateName, @RequestBody String template) {
+    return userNotificationService.updateNotificationTemplate(templateName, template);
+  }
+
+  @PostMapping(
+      value = "/notifications/template/{templateName}/reset",
+      produces = MediaType.APPLICATION_XHTML_XML_VALUE)
+  @Operation(
+      summary = "Reset a notification email-template",
+      description = "Reset a notification email-template to the default content")
```
RadovanTomik, line 58: according to REST conventions we should not use verbs in path, so maybe figure out another way to achieve the same result?

backend/src/main/java/eu/bbmri_eric/negotiator/notification/UserNotificationService.java
```diff
@@ -52,4 +52,27 @@ public interface UserNotificationService {
...
   /** Send out emails for all pending notifications. */
   void sendEmailsForNewNotifications();
+
+  /**
+   * Get a notification template.
+   *
+   * @param templateName the name of the template.
+   * @return the template.
+   */
+  String getNotificationTemplate(String templateName);
```
RadovanTomik, line 62: IHMo this violates the Interface Segregation Principle. Any good reason for it?

backend/src/main/java/eu/bbmri_eric/negotiator/notification/UserNotificationServiceImpl.java
```diff
@@ -169,6 +189,70 @@ public void notifyUsersAboutNewPost(Post post) {
...
+      String defaultTemplate = new String(inputStream.readAllBytes(), StandardCharsets.UTF_8);
+
+      String targetTemplatePath = thymeleafPrefix + templateName + thymeleafSuffix;
+      writeTemplateToFile(targetTemplatePath, defaultTemplate);
+
+      return defaultTemplate;
+
+    } catch (IOException e) {
+      log.error("Failed to reset notification template", e);
+      throw new RuntimeException("Failed to reset notification template", e);
```
RadovanTomik, line 252: imho this is mixing the business logic with retrieval logic. not sure the Service should care where the templates are stored

backend/src/test/java/eu/bbmri_eric/negotiator/integration/api/v3/NotificationControllerTest.java
```diff
@@ -62,4 +62,68 @@ void getNotifications_3new_Ok() throws Exception {
...
+  @Test
+  @WithUserDetails("admin")
+  void getNotificationTemplate_validTemplateName_Ok() throws Exception {
+    mockMvc
+        .perform(MockMvcRequestBuilders.get("/v3/notifications/template/footer"))
+        .andExpect(status().isOk())
+        .andExpect(content().contentType("application/xhtml+xml;charset=UTF-8"));
+  }
+
+  @Test
```
RadovanTomik, line 75: I dont see any test for validating the the templates are actually used in the emails

backend/src/main/resources/application-test.yaml
```diff
@@ -17,6 +17,8 @@ spring:
         jwt:
           jwk-set-uri: "https://testoidc"
           type: "JWT"
+  thymeleaf:
+    prefix: "file:../config/templates/"
```
RadovanTomik, line 21: what about production code?

docs/notifications.md
```diff
@@ -57,13 +57,18 @@ The email template uses variables to include dynamic content. These variables ar
...
+spring:
+  thymeleaf:
+    prefix: file:resources/templates/
 ```
- - **Hot Reload:** If using a development environment that supports hot reloading (like Spring Boot DevTools), you can take advantage of this feature to see changes immediately after saving the template files.
- - **Production Considerations:** While runtime updates are useful in development, it's recommended to enable caching in production to improve performance. Ensure templates are thoroughly tested before deploying updates in a production environment.
+Using the endpoint, you can update the email template without restarting the application. This is useful for making quick changes or tweaks to the template. 
+The content of the email template can be updated by sending a POST request to the `/api/v3/notifications/template/<template_name>` endpoint with the new template content in the request body. This endpoint is protected by the `ROLE_ADMIN` role. 
+To reset the template to the default content, send a POST request to the `/api/v3/notifications/template/<template_name>/reset` endpoint.
+
```
RadovanTomik, line 71: How would the user know the template names? Why not include some examples and what is allowed and what is not.

RadovanTomik (changes requested):
Please see the comments, generally I do not see how would this work in a production environment. Can this really work in a Docker container without any changes to the Dockerfile? Also why not add the UI part to this PR as well?

backend/src/main/java/eu/bbmri_eric/negotiator/common/configuration/security/HTTPRegistryConfigurer.java
```diff
@@ -59,6 +59,8 @@ public void configure(
         .authenticated()
         .requestMatchers(mvc.pattern("/v3/notifications"))
         .authenticated()
+        .requestMatchers(mvc.pattern("/v3/notifications/template/**"))
+        .hasRole("ADMIN")
```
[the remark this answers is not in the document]
jungwire: changes

RadovanTomik (changes requested):
we should move the email template editor to the admin console page, also we currently only support editing existing templates not creating new ones. Why is the HTML editor only for creating templates but not editing them? The section should be simple: a list/dropdown of existing templates and the editor

backend/src/main/resources/application-test.yaml
jungwire: changed it for production as well

backend/src/main/java/eu/bbmri_eric/negotiator/TemplateInitiationRunner.java
jungwire: split it in multiple functions

backend/src/main/java/eu/bbmri_eric/negotiator/notification/NotificationController.java
jungwire: Yes your are correct, the functionality was seperated from the contoller

backend/src/main/java/eu/bbmri_eric/negotiator/notification/UserNotificationServiceImpl.java
jungwire: this was separated

backend/src/main/java/eu/bbmri_eric/negotiator/notification/email/EmailTemplateController.java
```diff
@@ -0,0 +1,64 @@
...
+      summary = "Update a notification email-template",
+      description = "Update the content of a notification email-template by name")
+  @ResponseStatus(HttpStatus.OK)
+  public String updateNotificationTemplate(
+      @PathVariable String templateName, @RequestBody String template) {
+    return emailTemplateService.updateNotificationTemplate(templateName, template);
+  }
+
+  @PostMapping(
+      value = "/email-templates/{templateName}/reset",
```
RadovanTomik, line 55: we should still not use verbs in endpoint names

backend/src/main/java/eu/bbmri_eric/negotiator/notification/email/EmailTemplateServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+public class EmailTemplateServiceImpl implements EmailTemplateService {
+  private EmailTemplateRepository emailTemplateRepository;
+
+  public EmailTemplateServiceImpl(EmailTemplateRepository emailTemplateRepository) {
+    this.emailTemplateRepository = emailTemplateRepository;
+  }
+
+  @Override
+  public List<String> getAllNotificationTemplates() {
+    log.info("Getting all notification templates.");
```
RadovanTomik, line 21: is this level of logging necessary?

docs/notifications.md
```diff
@@ -57,13 +57,36 @@ The email template uses variables to include dynamic content. These variables ar
...
+  -d '<!DOCTYPE html>
+<html xmlns:th="http://www.thymeleaf.org" lang="en">
+    <head>
+        <title>Email notification</title>
+    </head>
+    <body>
+    Hello <span th:utext="${recipient.getName()}"></span>
+    </body>
+</html>'
 ```
```
RadovanTomik, line 88: we should update the documentation. No need to document the API endpoint when we have a UI, plus I would put it under admin documentation not dev.

RadovanTomik (changes requested):
please see comments, but otherwise seems fine

backend/src/main/java/eu/bbmri_eric/negotiator/notification/email/EmailTemplateController.java
```diff
@@ -0,0 +1,72 @@
...
+  @Operation(
+      summary = "Update a notification email-template",
+      description = "Update the content of a notification email-template by name")
+  @ResponseStatus(HttpStatus.OK)
+  public String updateNotificationTemplate(
+      @PathVariable String templateName, @RequestBody String template) {
+    return emailTemplateService.updateNotificationTemplate(templateName, template);
+  }
+
+  @PatchMapping(
```
RadovanTomik, line 56: not sure this is the best approach, patch is used to update a specific field of a Resource (this does not seem to be the case). Either we go back to /reset which breaks our commitment to keep the API RESTful, or we do /email-templates/name/operations which accepts a DTO that has only one attribute enum "value" that currently only supports RESET.

RadovanTomik (changes requested):

RadovanTomik (approved):

PR #720, opened 2025-04-09
feat: refactor form save draft

tmilost:
### Description:

### Checklist:


RadovanTomik (changes requested):
The responsivness is much better, but the form is saved after user clicks on next (meaning the first page of the form is not saved until then). It will not allow the user to click on next unless all required fields are filled. I would suggest to save as draft the moment user opens the page and bring back the save draft button

RadovanTomik (approved):

PR #759, opened 2025-05-12
Fix/file upload in information submission

tmilost:
### Description:

add attachment upload in submitRequiredInformation and after display attachments in form view modal
### Checklist:

- [ ] I have described the PR and added a meaningful title in the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format

### DIFF
--- a/frontend/src/components/modals/FormSubmissionModal.vue
+++ b/frontend/src/components/modals/FormSubmissionModal.vue
@@ -1,7 +1,7 @@
 <template>
   <div
     :id="id"
-    class="modal"
+    class="modal form-submission-modal"
     :class="{ fade: fade }"
     tabindex="-1"
     :aria-labelledby="`${id}Label`"
--- a/frontend/src/components/modals/FormViewModal.vue
+++ b/frontend/src/components/modals/FormViewModal.vue
@@ -20,10 +20,10 @@
             aria-label="Close"
           />
         </div>
-        <div class="modal-body text-left">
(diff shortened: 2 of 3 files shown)


RadovanTomik (changes requested):
works, but the information about submission is not updated, it should change from red to green on successful submission. PLease also handle attachment upload problems (what happens if the user tries to upload an unsupported file)

tmilost:
> Please also handle attachment upload problems (what happens if the user tries to upload an unsupported file)

can you give me example of problem because file attachment is already handled, try uploading exe file or bash
@RadovanTomik

RadovanTomik:
no its not, I can upload any file on mac and the API naturally returns 400. So please add handling for any response other than 200 and currently I don't see any attachment. Also please do not reload the whole page (Will take ages for large requests)
Screenshot 2025-05-14 at 10 39 29

RadovanTomik:
also the upload can fail because the file is too big so please really add error handling (if upload fails keep the modal open and just display the error so the user can correct it)

RadovanTomik:
successful attachment upload now works as expected. But when I try to upload a 6MB PDF I again see this as the attachment upload failed. According to the dev tools the HTTP request was cut short: 
Screenshot 2025-05-15 at 2 42 16
Screenshot 2025-05-15 at 2 43 41

RadovanTomik (approved):

PR #761, opened 2025-05-13
fix: disable file upload on every update of form data

tmilost:
### Description:

### Checklist:

- [ ] I have described the PR and added a meaningful title in the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format

### DIFF
--- a/frontend/src/components/NegotiationForm.vue
+++ b/frontend/src/components/NegotiationForm.vue
@@ -546,30 +546,32 @@ async function getValueSet(id) {
 
 async function saveDraft(step) {
   if ((!props.isEditForm || currentStatus.value === 'DRAFT') && currentSectionModified.value) {
-    await saveNegotiation(true, step)
+    await saveNegotiation(true, step, true)
     currentSectionModified.value = false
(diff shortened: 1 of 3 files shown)


RadovanTomik (changes requested):
if I upload an attachment and save the negotiation as draft the attachment is lost

RadovanTomik (changes requested):
see comments

RadovanTomik (dismissed):
The error message for failed attachment upload disappears to quickly, the user doesn't have a chance to read the error message.
If I replace the attachment I have uploaded I can then still see both in the NEgotiation Page.
The attachment download from the form details doesn't work as it tries to fetch http://localhost:8080/api/v3/attachments/[object%20Object]

tmilost:
- fixed bug with attachment download
- I don't see a way to delete previous attachments that are connected to the negotiation, there is no api
- for notifications bugs i created ticket that i will solve separately https://github.com/BBMRI-ERIC/negotiator/issues/768
@RadovanTomik

PR #845, opened 2025-08-14
feat: notifying about resource registration

ChrisiSailer:
### Description:

A user is notified upon being registered as a representative for a resource.

### Checklist:

### DIFF
--- a/backend/src/main/java/eu/bbmri_eric/negotiator/notification/internal/RepresentativeAddedHandler.java
+++ b/backend/src/main/java/eu/bbmri_eric/negotiator/notification/internal/RepresentativeAddedHandler.java
@@ -9,6 +9,9 @@ import eu.bbmri_eric.negotiator.user.PersonRepository;
 import eu.bbmri_eric.negotiator.user.RepresentativeAddedEvent;
 import jakarta.transaction.Transactional;
 import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import lombok.extern.apachecommons.CommonsLog;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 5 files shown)


backend/src/main/java/eu/bbmri_eric/negotiator/user/PersonServiceImpl.java
```diff
@@ -135,6 +135,7 @@ public void assignAsRepresentativeForResource(Long representativeId, Long resour
       eventPublisher.publishEvent(
           new FirstRepresentativeEvent(this, resource.getId(), resource.getSourceId()));
     }
+    eventPublisher.publishEvent(new RepresentativeAddedEvent(this, resourceId, representativeId));
```
RadovanTomik, line 138: I would put this after the changes are persisted, what if there is an error during persistence?

backend/src/main/java/eu/bbmri_eric/negotiator/notification/internal/RepresentativeAddedHandler.java
```diff
@@ -0,0 +1,60 @@
...
+  @Override
+  @Transactional
+  public void notify(RepresentativeAddedEvent event) {
+    Long representativeId = event.getRepresentativeId();
+    Long resourceId = event.getResourceId();
+
+    Person representative = personRepository.findById(representativeId).orElse(null);
+    Resource resource = resourceRepository.findById(resourceId).orElse(null);
+
+    if (representative != null && resource != null) {
```
RadovanTomik, line 48: no logging if something went wrong and one of them is null?

backend/src/main/java/eu/bbmri_eric/negotiator/notification/internal/RepresentativeAddedHandler.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.List;
+import lombok.extern.apachecommons.CommonsLog;
+import org.springframework.stereotype.Component;
+
+@Component
+@CommonsLog
+class RepresentativeAddedHandler implements NotificationStrategy<RepresentativeAddedEvent> {
+
+  public static final String TITLE = "Added as Representative";
+  public static final String BODY = "You have been added as a representative for the resource: ";
```
RadovanTomik, line 20: Can you attach a screenshot of what such an email looks like?

RadovanTomik (changes requested):

RadovanTomik:
@ChrisiSailer if I understood the code correctly, if a person was added as a representative to 100 resources, they would get a hundred emails, correct? If so then I think we should come up with a different solution. The goal is to show representatives what resources they represent, so the story would go like" As a resource representative, I would like to see what resources I represent to have an overview of my responsibilities". Can you suggest something?

PR #860, opened 2025-08-25
feat: return to template list after temple is updated

tmilost:
### Description:

Return to template list after temple is updated

### Checklist:


RadovanTomik:
should be fixed in #859

PR #872, opened 2025-09-03
feat: add represented resources page

ChrisiSailer:
### Description:

Through these changes a representative of a resources should be able to view all represented resources grouped by the organization.

### Checklist:


ChrisiSailer:
Below are screenshots illustrating the implemented resource page:

Resource page:

Resource_Page

Resource page with filtering:

Resource_Page_Filtering

Resource page withdrawn only:

Resource_Page_Withdrawn

RadovanTomik (changes requested):
please see my comments. I would still say it would be better to just fetch all Resources via existing endpoints rather than adding a new one.

PR #904, opened 2025-09-25
903 bug fix error page on unauthorized views

tmilost:
### Description:

As a researcher I have no business going into the governance page. However this error message seem weird at best. Currently there is a method in index for verifying if as a user I'm allowed to visit a certain page. This does not make sense as the error page should be thrown up automatically if the API returns 401/403.

### Checklist:

- [] I have updated the documentation in all relevant places (Javadoc, Swagger, MDs...)

### DIFF
--- a/frontend/src/router/index.js
+++ b/frontend/src/router/index.js
@@ -17,6 +17,25 @@ import hasUser from '@/middlewares/hasUser.js'
 import middlewarePipeline from '@/middlewares/middleware-pipeline.js'
 import { useNotificationsStore } from '@/store/notifications'
 
+async function isAllowedToAccessCombinded(roles) {
+  const userStore = useUserStore()
+  const notifications = useNotificationsStore()
+
+  if (Object.keys(userStore.userInfo).length === 0) {
+    await userStore.retrieveUser()
+  }
+
+  for (const role of roles) {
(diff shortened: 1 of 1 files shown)


frontend/src/router/index.js
```diff
@@ -78,17 +81,27 @@ const router = createRouter({
...
       component: GovernancePage,
       meta: { isPublic: false, middleware: [hasUser] },
-      beforeEnter: async () => {
-        return await isAllowedToAccess(ROLES.REPRESENTATIVE) || await isAllowedToAccess(ROLES.ADMINISTRATOR)
+      beforeEnter: async (to, from, next) => {
+        const isAllowed = await isAllowedToAccess([ROLES.REPRESENTATIVE, ROLES.ADMINISTRATOR])
+        if (isAllowed) {
+          next() // Allow access
+        } else {
+          next('/error-page') // Redirect to the home page or another route
```
RadovanTomik, line 103: cant this duplications be extracted into a helper method? If to me this seems like the same logic: IF not allowed, THEN redirect to error page. Which seems like something you want to be done whenever a page is not accessible to a user

RadovanTomik (changes requested):

konradlang (approved):

tmilost-bbmri-eric (approved):

Issue #1009, opened 2026-01-28
[FEATURE] Comprehensive test environment setup


konradlang:
**Brief description.**
Additional setup for development including setup for testing.

**Desired functionality acceptance criteria**
When test environment is deployed this includes already negotiations and example setup for a variety of use cases.

PR #1025, opened 2026-02-05
feat: add negotiation sample data

stetsche:
### Description:

Adding sample negotiations to the initial dataset in test setups. Includes negotiations with different states and messages.
Addresses #1009 .

### Checklist:


RadovanTomik:
hey, @stetsche I would suggest rewriting the test data script to make it more idempotent. The thing is that Flyway detects that the checksum of the migration file has changed so it tries to rerun the migration again, which causes the DB migration step in CI to fail. For example, we could switch plain INSERT statements to INSERT … ON CONFLICT DO NOTHING (or an equivalent upsert) and guard DDL changes with IF NOT EXISTS, so the migration can be safely re-executed.

Issue #1026, opened 2026-02-09
[BUG] Webhook SSL bypass can affect other connections


stetsche:
**Describe the bug**
The current SSL bypass implementation disables all certificate verification using a custom `X509TrustManager` that accepts any certificate.

**What happens**
The insecure SSL certificate and hostname validation set statically as the default, which could affect other connections.
```
// WebhookServiceImpl.java
HttpsURLConnection.setDefaultSSLSocketFactory(sslContext.getSocketFactory());
HttpsURLConnection.setDefaultHostnameVerifier((hostname, session) -> true);
```
**What should happen**
Insecure webhook deliveries (for testing) should not affect other connections which should be handled in a secure context.

PR #1037, opened 2026-02-11
fix: safer ssl verification bypass

stetsche:
### Description:

Addressing #1026 . 
- Configured rest template with custom ssl context, not applying as default
- Reuse webhook rest templates and enable configuration options to make it more testable
- Add tests to verify ssl bypass is working

### Checklist:


backend/src/test/java/eu/bbmri_eric/negotiator/integration/service/WebhookDeliverySslIntegrationTest.java
```diff
@@ -0,0 +1,152 @@
...
+                aResponse()
+                    .withStatus(200)
+                    .withHeader("Content-Type", MediaType.APPLICATION_JSON_VALUE)
+                    .withBody("{\"status\":\"received\"}")));
+
+    String payload = "{\"data\":\"success\"}";
+
+    DeliveryDTO delivery = webhookService.deliver(payload, webhook.getId());
+
+    assertEquals(500, delivery.getHttpStatusCode());
```
RadovanTomik: IHMO this would benefit from an assertion that the error message contains metion of SSL/certificate, that the error message it tied to ssl validation

PR #1126, opened 2026-03-27
feat: push governance to lsaai

svituz:
### Description:

This PR adds the possibility of managing representatives directly, using Perun's LS-AAI.
When configured to use Perun, it allows adding members to the groups of representatives of collections in Perun. The interaction with Perun is done via the [Perun's API](https://perun-aai.org/documentation/technical-documentation/rpc-api/index.html). To call Perun's API, the frontend uses the same JWT used for the Negotiator.

For instances that don't manage representatives using Perun, it is possible to keep the old implementation using the Negotiator APIs. Indeed, the type of governance client is configurable via a specific setting: it can be `NEGOTIATOR` or `PERUN`. In the first case, the Governance page will act by sending requests to the Negotiator APIs, as it is now.

In the second case, the process to assign a user as representative of a resource is the following:
* Log in as a Network Manager to the `Governance` page. 
* The page will show the hierarchy of Organizations and Resources. To do that, it:
   *  Gets all the groups from Perun (Perun returns only the groups the user is allowed to see)
   *  Filters the groups associated with Biobanks and Collections used to assign representatives, and creates the hierarchy
   *  Query the Negotiator for Organizations to get the missing information to show on the page. NB: the external source is in an attribute of Perun's group.
*  When the user clicks an Organization, the list of resources is shown. Now the Negotiator is queried for information about the resources of the Organization.
* When the user selects a resource for representative assignment, the page shows the modal with the list of current representatives obtained from Perun
* When the user searches for members, the modal calls Perun for members (NB: Perun has an overall search string where the user can put the Name or the Email of the user, so two separate fields are not useful here).
* When the user selects one or more members to be added or removed as representative, the page will call Perun to add/remove the users from the representatives' group associated with the resource

In case the logged-in user is a representative, the Governance page will show the represented resources by calling the Negotiator API

To test this PR you need to:
* configure values for Perun API in config/governance
* have an account for a Network Manager authorization in Perun
* have an account for a representative.

The steps to test are:
* in Perun, add an account as manager of a Network Manager
* in the Negotiator DB add that account as network manager for the same network. This is to simulate synchornzization script
* log in the negotiator with that account
* go to governance page and add the representative user to one of the resources
* modify the negotiator database manually to add the same user as representative of the resource in the negotiator (this will simulate the synchronization)
* login in the negotiator as the representative, go in the Governance page and check that the reource is shown

* now login again as the network manager and remove the representative
* login again as the representative and access the page. An error message should be displayed, that it was not possible to load represented resource. Ths is because perun's won't return the groups now

### Checklist:

- [X] I have performed a self-review of my code
- [X] I have made my code as simple as possible
- [X] I have removed all commented code
- [X] I have described the PR and added a meaningful title in the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
**If applicable to this PR:**


frontend/src/components/NavigationBar.vue
```diff
@@ -56,13 +56,11 @@
               {{ $t('navbar.biobanker') }}
             </router-link>
           </li>
-          <li v-if="isRepresentative || isAdmin" class="nav-item v-step-13">
+          <li v-if="isRepresentative || isGovernanceManager" class="nav-item v-step-13">
```
stetsche, line 59: Should we not keep `|| isAdmin` as well?

frontend/src/components/NavigationBar.vue
```diff
@@ -56,13 +56,11 @@
...
             <router-link
               :style="{
                 color:
-                  $route.path === '/governance' ||
-                  $route.params.userRole === 'ROLE_REPRESENTATIVE' ||
-                  $route.params.userRole === 'ROLE_ADMIN'
+                  $route.path === '/governance' || isRepresentative || isGovernanceManager
```
stetsche, line 63: Do we need `|| isRepresentative || isGovernanceManager`? It is already checked on the parent and  seems to only affect coloring.

frontend/src/utils/governance/groupsManager.js
```diff
@@ -0,0 +1,217 @@
...
+
+const getNegotiatorResourceIdFromRepresentativeGroup = (group) => {
+  return getValueForAttribute(group, RESOURCE_ID_ATTR).replaceAll('.', ':')
+}
+
+const getNegotiatorOrganizationIdFromRepresentativeGroup = (group) => {
+  return getValueForAttribute(group, ORGANIZATION_ID_ATTR).replaceAll('.', ':')
+}
+
+function OrganizationRepresentativeGroup(perunGroupId) {
```
stetsche, line 32: It would be nice if this could be in composables as something like useOrganizationRepresentativeGroup(perunGroupId) and just return an object with all the relevant refs. Also wrapping the returned object in useMemorize.
This also applies to the other functions in this file with similar structure.
If you opt for keeping this as a util, you could also use regular variables rather then vue `ref`s here, maybe also utilizing `class` setting defaults in the constructor.

frontend/src/utils/governance/index.js
```diff
@@ -0,0 +1,19 @@
...
+  if (getGovernanceClient.client == null) {
+    switch (governanceSettings.client) {
+      case 'NEGOTIATOR':
+        getGovernanceClient.client = NegotiatorClient()
+        break
+      case 'PERUN':
+        getGovernanceClient.client = PerunClient()
+        break
+      default:
+        throw new Error(`Unknown provider type: ${governanceSettings.client}`)
```
stetsche, line 15:
```suggestion
        throw new Error(`Unknown governance provider: ${governanceSettings.client}`)
```

frontend/src/utils/governance/index.js
```diff
@@ -0,0 +1,19 @@
+import governanceSettings from '@/config/governance'
+import { NegotiatorClient } from './negotiatorClient'
+import { PerunClient } from './perunClient'
+
+export function getGovernanceClient() {
```
stetsche, line 5: I think most of the added `utils` code should be in `composables` using naming like `useGovernanceClient`. To me, the current structure is hiding stateful dependencies the user store.

stetsche (changes requested):
Looks pretty good, I added just added a few suggestions to the structure in the `utils` code and a doubt regarding CORS. Please also check the Copilot comments.

stetsche:
Converting to draft for now, please mark as ready when we should review again.

PR #1179, opened 2026-06-02
fix: fix docker-compose postgres healthcheck command format

### LINKED ISSUE #1171: [BUG] Postgres health check in docker compose is broken
**Describe the bug**
The docker health check in `compose.yml` for the postgres (negotiator-db) container is not working and kills the container. The issue seems to be two-fold. For one, we don't pass the user argument, so `root` may be used as the postgres user. And also, we use `CMD-SHELL`, which should receive only one command string rather then an array of command parts. we can keep the current syntax using `CMD` rather then `CMD-SHELL`.
 
**How to reproduce**
`docker compse up`, check logs.

**What happens**
`FATAL:  role "root" does not exist` every 5s

**What should happen**
postgres container (negotiator-db) stays healty and is up and running

tmilost:
### Description:

fix(docker): correct postgres healthcheck syntax

Fixes the `FATAL: role "root" does not exist` error caused by an 
incorrectly formatted `CMD-SHELL` array. 

By wrapping the entire `pg_isready` command into a single string argument, 
Docker Compose now correctly passes the user flags (`-U negotiator`) to 
the shell execution environment instead of dropping them and defaulting to root.

### Checklist:

**If applicable to this PR:**

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -38,7 +38,7 @@ services:
       - POSTGRES_PASSWORD=negotiator
       - POSTGRES_DB=negotiator
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U negotiator -d negotiator"]
+      test: ["CMD-SHELL", "pg_isready -U $${POSTGRES_USER} -d $${POSTGRES_DB}"]
       interval: 5s
       timeout: 60s
       retries: 5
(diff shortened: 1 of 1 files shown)


stetsche:
Closing as duplicate

tmilost:
duplicate https://github.com/BBMRI-ERIC/negotiator/pull/1172

PR #1189, opened 2026-06-09
feat(forms): reuse shared capitalizeAllWords for access form labels

### LINKED ISSUE #1079: [FEATURE] Capitalize From field name
**Brief description.**
We should capitalise form field names for better readability

tmilost:
### Description:

Description:

- Replaced the local label-capitalization logic with the shared capitalizeAllWords helper.
- Updated access form field label rendering to use the common utility.
- Removed duplicate formatting code from the component to improve consistency and maintainability.
- This ensures form field names are capitalized uniformly across the app.

### Checklist:

**If applicable to this PR:**

### DIFF
--- a/frontend/src/composables/utils.js
+++ b/frontend/src/composables/utils.js
@@ -2,6 +2,15 @@ export function capitalizeFirstLetter(val) {
   return String(val).charAt(0).toUpperCase() + String(val).slice(1)
 }
 
+export function capitalizeAllWords(val) {
+  if (!val) return ''
+  return val
+    .split(/[\s_-]+/)
+    .filter(Boolean) // remove empty strings from edge cases
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1).toLowerCase())
+    .join(' ')
+}
+
 export function transformStatus(status) {
   if (status === 'SUBMITTED') {
     return 'UNDER REVIEW'
--- a/frontend/src/composables/utils.js
+++ b/frontend/src/composables/utils.js
@@ -3,8 +3,9 @@ export function capitalizeFirstLetter(val) {
 }
 
 export function capitalizeAllWords(val) {
-  if (!val) return ''
-  return val
+  const str = String(val ?? '').trim()
+  if (!str) return ''
+  return str
     .split(/[\s_-]+/)
     .filter(Boolean) // remove empty strings from edge cases
     .map((word) => word.charAt(0).toUpperCase() + word.slice(1).toLowerCase())
--- a/frontend/src/components/OrganizationListItem.vue
+++ b/frontend/src/components/OrganizationListItem.vue
@@ -188,9 +188,7 @@
                   <span v-if="resource.accessForm">
                     {{ resource.accessForm.name }}
                   </span>
-                  <span v-else class="text-warning">
-                    None assigned
-                  </span>
+                  <span v-else class="text-warning"> None assigned </span>
                 </small>
               </div>
             </div>
--- a/frontend/src/components/form-components/AccessFormSection.vue
+++ b/frontend/src/components/form-components/AccessFormSection.vue
@@ -2,7 +2,7 @@
   <div class="access-form-section d-flex flex-column">
     <div class="section mb-5">
       <h2 class="fw-bold">
-        {{ accessFormWithPayloadSection.label }}
+        {{ capitalizeAllWords(accessFormWithPayloadSection.label) }}
       </h2>
       <p class="mb-0">{{ accessFormWithPayloadSection.description }}</p>
     </div>
@@ -23,7 +23,7 @@
       "
     >
       <label class="form-label" :class="{ required: element.required }">
-        {{ element.label }}
+        {{ capitalizeAllWords(element.label) }}
       </label>
       <span v-if="element.description" class="ms-2 text-muted">
         <i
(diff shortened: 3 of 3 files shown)


frontend/src/composables/utils.js
```diff
@@ -2,6 +2,15 @@ export function capitalizeFirstLetter(val) {
...
 }
 
+export function capitalizeAllWords(val) {
+  if (!val) return ''
+  return val
+    .split(/[\s_-]+/)
+    .filter(Boolean) // remove empty strings from edge cases
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1).toLowerCase())
+    .join(' ')
+}
```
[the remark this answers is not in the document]
tmilost: resolved

frontend/src/composables/utils.js
```diff
@@ -2,6 +2,16 @@ export function capitalizeFirstLetter(val) {
   return String(val).charAt(0).toUpperCase() + String(val).slice(1)
 }
 
+export function capitalizeAllWords(val) {
```
stetsche, line 5: @tmilost @konradlang Should we preserve the capital letters of acronyms? With this code, they are title-cased. Also, do we need to split by `-` as well? Just thinking, if we only use `_` of ` ` for splitting words, we could support the use of `-` within the labels.
```
> function capitalizeAllWords(val) {
|   const str = String(val ?? '').trim()
|   if (!str) return ''
|   return str
|     .split(/[\s_-]+/)
|     .filter(Boolean) // remove empty strings from edge cases
|     .map((word) => word.charAt(0).toUpperCase() + word.slice(1).toLowerCase())
|     .join(' ')
| }
undefined
> capitalizeAllWords("BBMRI-ERIC")
'Bbmri Eric'
```

konradlang: Good point, capital letters of acronyms should be preserved.
Would also not do a split for `-`

stetsche (changes requested):

stetsche:
Looking at this PR again, the access form element labels are set by the admin (or in flyway scripts) in data. Should we rather update the data than forcing the format? @tmilost @konradlang

frontend/src/composables/utils.js
tmilost: Good catch, both — fixed in capitalizeAllWords (frontend/src/composables/utils.js):

Acronyms are now preserved: if a token is already fully uppercase (e.g. ERIC, ID), we skip the lowercase pass instead of mangling it to Eric.
- is no longer a hard word separator alongside _/space. It's now treated as an internal delimiter — each side gets capitalized independently but the hyphen itself is kept literal in the output, instead of being collapsed into a space.

capitalizeAllWords("BBMRI-ERIC")      // "BBMRI-ERIC"  (was "Bbmri Eric")
capitalizeAllWords("co-branded_service") // "Co-Branded Service"
capitalizeAllWords("access_form_section") // "Access Form Section" (unchanged)

tmilost:
> Looking at this PR again, the access form element labels are set by the admin (or in flyway scripts) in data. Should we rather update the data than forcing the format? @tmilost @konradlang

Since labels are admin-editable at runtime (/v3/elements, /v3/sections), not just flyway-seeded, I don't think fixing the data alone solves it — new admin input can still be inconsistently cased. I'd keep the display-side formatting as a safety net regardless of what we do with existing seed data.

stetsche:
IMO that's a feature, not a bug. Admin configures, how the form label should be cased and if it has dashes, how acronyms are spelled, etc. The admin can change it at any time. This would also affect existing form fields:
image
image

tmilost:
> IMO that's a feature, not a bug. Admin configures, how the form label should be cased and if it has dashes, how acronyms are spelled, etc. The admin can change it at any time. This would also affect existing form fields: image image

@konradlang  can you decide in wich direction to go, to force format (what is in this pr commited) or leave users to change format is acces form settings

konradlang:
> > IMO that's a feature, not a bug. Admin configures, how the form label should be cased and if it has dashes, how acronyms are spelled, etc. The admin can change it at any time. This would also affect existing form fields: image image
> 
> @konradlang can you decide in wich direction to go, to force format (what is in this pr commited) or leave users to change format is acces form settings

I'm voting for auto capitalization. Worst case we can still add another feature flag to allow this to be controlled  deployment wide.

stetsche:
This fails on following label used in the Isidore instance:
```
> capitalizeAllWords("Short CV(s) of the research group members/PI")
'Short Cv(s) Of The Research Group Members/pi'
```
Acronym PI get's lowercased. Also, CV should be preserved I think.

tmilost:
> Short CV(s) of the research group members/PI

tested
`{{ capitalizeAllWords('resource_id') }} -> Resource Id`
`{{ capitalizeAllWords('access conditions') }} -> Access Conditions`
`{{ capitalizeAllWords('ERIC') }} -> ERIC`
`{{ capitalizeAllWords('project-name') }} -> Project-Name`
`{{ capitalizeAllWords('first_name last_name') }} -> First Name Last Name`
`{{ capitalizeAllWords('Short CV(s) of the research group members/PI') }} -> Short CV(s) Of The Research Group Members/PI`
`{{ capitalizeAllWords('') }} -> ''`
`{{ capitalizeAllWords(null) }} -> ''`

stetsche (approved):
Looks good now. One more thing I would add, but leave it up to you, is to apply this formatting in the Access Form Elements table in the admin interface, so the final format is visible there as well.

PR #1222, opened 2026-06-29
refactor(pdf): replace Flying Saucer with iText pdfHTML

### LINKED ISSUE #1159: [FEATURE] Upgade flying-saucer-pdf to 10.x
**Brief description.**
An flying-saucer-pdf version upgrade from 9.x to 10.x will require dependecy or code changes, when just changing the version, the backend does not compile:
```
 COMPILATION ERROR : 
Error:  /home/runner/work/negotiator/negotiator/backend/src/main/java/eu/bbmri_eric/negotiator/attachment/DocConverter.java:[3,24] package com.lowagie.text does not exist
Error:  /home/runner/work/negotiator/negotiator/backend/src/main/java/eu/bbmri_eric/negotiator/attachment/DocConverter.java:[4,24] package com.lowagie.text does not exist
Error:  /home/runner/work/negotiator/negotiator/backend/src/main/java/eu/bbmri_eric/negotiator/attachment/DocConverter.java:[5,28] package com.lowagie.text.pdf does not exist
Error:  /home/runner/work/negotiator/negotiator/backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/pdf/NegotiationPdfServiceImpl.java:[6,28] package com.lowagie.text.pdf does not exist
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.14.1:compile (default-compile) on project negotiator: Compilation failure: Compilation failure: 
Error:  /home/runner/work/negotiator/negotiator/backend/src/main/java/eu/bbmri_eric/negotiator/attachment/DocConverter.java:[3,24] package com.lowagie.text does not exist
Error:  /home/runner/work/negotiator/negotiator/backend/src/main/java/eu/bbmri_eric/negotiator/attachment/DocConverter.java:[4,24] package com.lowagie.text does not exist
Error:  /home/runner/work/negotiator/negotiator/backend/src/main/java/eu/bbmri_eric/negotiator/attachment/DocConverter.java:[5,28] package com.lowagie.text.pdf does not exist
Error:  /home/runner/work/negotiator/negotiator/backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/pdf/NegotiationPdfServiceImpl.java:[6,28] package com.lowagie.text.pdf does not 
```
Failed dependabot upgrade:
https://github.com/BBMRI-ERIC/negotiator/actions/runs/25673920789/job/75366377927?pr=1012

- flying-saucer-pdf version is latest 10.x
- backend compiles

fuchs-an:
Flying Saucer 10.x no longer bundles the iText 2.x (`com.lowagie.text`) classes, so bumping the version alone does not compile (#1159). Rather than pinning the unmaintained iText 2.1.7 (2009) as an explicit dependency, this PR drops Flying Saucer entirely and renders PDFs with iText, which provides HTML-to-PDF rendering itself via pdfHTML.

### Description:

- `backend/pom.xml`: remove `org.xhtmlrenderer:flying-saucer-pdf`, add `com.itextpdf:html2pdf` (6.3.3) and `kernel` / `layout` / `io` (9.7.0). 9.7.0 is exactly what pdfHTML 6.3.3 resolves to, so all iText modules converge on one version. iText is dual-licensed AGPL/commercial, which fits this project.
- `NegotiationPdfServiceImpl`: render the Thymeleaf summary through `HtmlConverter` with a `DefaultFontProvider` instead of `ITextRenderer`.
- `DocConverter`: rewritten on the iText layout API (`PdfDocument`, `Document`, `Paragraph`); the document is managed with try-with-resources, which keeps the error handling as it was before the migration.
- `NegotiationPdfGenerationTest`: regression test for the "Download PDF with attachments" export — it attaches a real docx and asserts the merged export has more pages than the summary-only export.

Closes #1159

### Checklist:

**If applicable to this PR:**

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -414,7 +414,7 @@
         <dependency>
             <groupId>com.lowagie</groupId>
             <artifactId>itext</artifactId>
-            <version>2.1.7</version>
+            <version>${itext.version}</version>
         </dependency>
 
         <!-- Development -->
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -29,7 +29,6 @@
         <commons-math3.version>3.6.1</commons-math3.version>
         <commons-validator.version>1.10.1</commons-validator.version>
         <docx4j.version>11.5.14</docx4j.version>
-        <flying-saucer.version>10.0.2</flying-saucer.version>
         <gson.version>2.12.1</gson.version>
         <httpcomponents.version>5.5.1</httpcomponents.version>
         <hypersistence.version>3.9.10</hypersistence.version>
@@ -51,7 +50,8 @@
         <poi.version>5.4.1</poi.version>
         <postgresql.version>42.7.11</postgresql.version>
         <reactor.version>3.7.3</reactor.version>
-        <itext.version>2.1.7</itext.version>
+        <itext.version>9.6.0</itext.version>
+        <itext.html2pdf.version>6.3.2</itext.html2pdf.version>
         <standardwebhooks.version>1.1.1</standardwebhooks.version>
         <spring-hateoas.version>2.5.1</spring-hateoas.version>
         <springdoc.version>2.8.9</springdoc.version>
@@ -408,13 +408,23 @@
             <version>${moxy.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.xhtmlrenderer</groupId>
-            <artifactId>flying-saucer-pdf</artifactId>
-            <version>${flying-saucer.version}</version>
+            <groupId>com.itextpdf</groupId>
+            <artifactId>html2pdf</artifactId>
+            <version>${itext.html2pdf.version}</version>
         </dependency>
         <dependency>
-            <groupId>com.lowagie</groupId>
-            <artifactId>itext</artifactId>
+            <groupId>com.itextpdf</groupId>
+            <artifactId>kernel</artifactId>
+            <version>${itext.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.itextpdf</groupId>
+            <artifactId>layout</artifactId>
+            <version>${itext.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.itextpdf</groupId>
+            <artifactId>io</artifactId>
             <version>${itext.version}</version>
         </dependency>
 
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -29,7 +29,7 @@
         <commons-math3.version>3.6.1</commons-math3.version>
         <commons-validator.version>1.10.1</commons-validator.version>
         <docx4j.version>11.5.14</docx4j.version>
-        <flying-saucer.version>9.5.1</flying-saucer.version>
+        <flying-saucer.version>10.0.2</flying-saucer.version>
         <gson.version>2.12.1</gson.version>
         <httpcomponents.version>5.5.1</httpcomponents.version>
         <hypersistence.version>3.9.10</hypersistence.version>
@@ -410,6 +410,11 @@
             <groupId>org.xhtmlrenderer</groupId>
             <artifactId>flying-saucer-pdf</artifactId>
             <version>${flying-saucer.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.lowagie</groupId>
+            <artifactId>itext</artifactId>
+            <version>2.1.7</version>
         </dependency>
 
         <!-- Development -->
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -51,6 +51,7 @@
         <poi.version>5.4.1</poi.version>
         <postgresql.version>42.7.11</postgresql.version>
         <reactor.version>3.7.3</reactor.version>
+        <itext.version>2.1.17</itext.version>
         <standardwebhooks.version>1.1.1</standardwebhooks.version>
         <spring-hateoas.version>2.5.1</spring-hateoas.version>
         <springdoc.version>2.8.9</springdoc.version>
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -51,7 +51,7 @@
         <poi.version>5.4.1</poi.version>
         <postgresql.version>42.7.11</postgresql.version>
         <reactor.version>3.7.3</reactor.version>
-        <itext.version>2.1.17</itext.version>
+        <itext.version>2.1.7</itext.version>
         <standardwebhooks.version>1.1.1</standardwebhooks.version>
         <spring-hateoas.version>2.5.1</spring-hateoas.version>
         <springdoc.version>2.8.9</springdoc.version>
--- a/backend/src/main/java/eu/bbmri_eric/negotiator/attachment/DocConverter.java
+++ b/backend/src/main/java/eu/bbmri_eric/negotiator/attachment/DocConverter.java
@@ -1,8 +1,9 @@
 package eu.bbmri_eric.negotiator.attachment;
 
-import com.lowagie.text.Document;
-import com.lowagie.text.Paragraph;
-import com.lowagie.text.pdf.PdfWriter;
+import com.itextpdf.kernel.pdf.PdfDocument;
+import com.itextpdf.kernel.pdf.PdfWriter;
+import com.itextpdf.layout.Document;
+import com.itextpdf.layout.element.Paragraph;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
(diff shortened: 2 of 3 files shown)


fuchs-an:
Right now, this is a band-aid; This fix only imports iText 2.1.7 manually. In my opinion we should update to a newer version (7+), but this will require some more extensive changes to NegotiationPdfServiceImpl.java and DocConverter.java. 

iText 2.1.7 is from 2009, and was previously implicitly imported by flying-saucer-pdf < 10. 

I would suggest to not accept this until this is reworked.

backend/src/main/java/eu/bbmri_eric/negotiator/attachment/DocConverter.java
```diff
@@ -49,9 +49,15 @@ public byte[] convertToPdf(byte[] docBytes) throws IOException {
...
       return result;
+    } catch (Exception e) {
+      throw new IOException("Error converting DOC to PDF", e);
     } finally {
       if (pdfDoc != null && pdfDoc.isOpen()) {
-        pdfDoc.close();
+        try {
+          pdfDoc.close();
+        } catch (Exception e) {
+          log.warn("Error closing PDF document", e);
```
stetsche, line 60: Let's catch the specific `Exception` we want to handle. Also use error level:
```suggestion
          log.error("Could no close PDF document", e);
```

backend/src/main/java/eu/bbmri_eric/negotiator/attachment/DocConverter.java
```diff
@@ -49,9 +49,15 @@ public byte[] convertToPdf(byte[] docBytes) throws IOException {
       byte[] result = pdfOutputStream.toByteArray();
       log.debug("Successfully converted DOC to PDF, output size: " + result.length);
       return result;
+    } catch (Exception e) {
```
stetsche, line 53: Let's catch the specific Exception we want to handle.

backend/pom.xml
```diff
@@ -29,7 +29,7 @@
         <commons-math3.version>3.6.1</commons-math3.version>
         <commons-validator.version>1.10.1</commons-validator.version>
         <docx4j.version>11.5.14</docx4j.version>
-        <flying-saucer.version>9.5.1</flying-saucer.version>
+        <flying-saucer.version>10.0.2</flying-saucer.version>
```
stetsche: Could we use the latest?
```suggestion
        <flying-saucer.version>10.3.0</flying-saucer.version>
```

fuchs-an:
The previous commit changed everything to use the newest iText - completely removing flying-saucer-pdf, as iText already has built-in functionality for rendering PDFs. This passes automated checks, but should be tested manually before acceptance.

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/pdf/NegotiationPdfServiceImpl.java
```diff
@@ -187,13 +187,13 @@ private byte[] renderPdf(String html) {
...
-          .getFontResolver()
-          .addFont(fontUrl.toString(), BaseFont.IDENTITY_H, BaseFont.EMBEDDED);
-      renderer.setDocumentFromString(html);
-      renderer.layout();
-      renderer.createPDF(outputStream);
+      FontProvider fontProvider = new DefaultFontProvider(false, false, false);
+      fontProvider.addFont(fontUrl.toString());
+
+      com.itextpdf.html2pdf.HtmlConverter.convertToPdf(
+          html,
```
stetsche, line 194: Please import `HtmlConverter`

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/pdf/NegotiationPdfServiceImpl.java
```diff
@@ -187,13 +187,13 @@ private byte[] renderPdf(String html) {
...
+          outputStream,
+          new com.itextpdf.html2pdf.ConverterProperties().setFontProvider(fontProvider));
```
stetsche, line 196: Please import `ConverterProperties`

backend/pom.xml
```diff
@@ -51,6 +50,8 @@
         <poi.version>5.4.1</poi.version>
         <postgresql.version>42.7.11</postgresql.version>
         <reactor.version>3.7.3</reactor.version>
+        <itext.version>9.6.0</itext.version>
+        <itext.html2pdf.version>6.3.2</itext.html2pdf.version>
```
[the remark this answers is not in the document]
stetsche: As we use AGPL i think the itext license model works fine for us. A BOM would be interesting, but I am not sure if such exists

stetsche (changes requested):
Nice work! I lightly tested this locally and compared with current master. Adding a simple docx file as attachment to the negotiation and `Download PDF with attachments`, master broke, and your version was working fine 😄 .
I found a small regression though where the "PROJECT" header now is not on the same page as the project description. Also the font has changed, but that's probably just fine.

I am attaching PDF exports (before+after), trace of the described bug, and the docx i used for context.

It would be nice if we could add a regression test for the export failure when exporting with attachment, but not a blocker. 

Regarding the code review, my comments assume that there was a new exception added, that now must be handled, if not we probably can keep the error handling as it was before.

fuchs-an:
Merged current master into the branch (the conflict was in the same `pom.xml` dependency block where master dropped EclipseLink MOXy) and worked through the review feedback:

- **Imports** — `HtmlConverter` and `ConverterProperties` are imported now instead of being fully qualified (@stetsche).
- **Error handling in `DocConverter`** — no new checked exception appears with the iText layout API, so, as you suggested, the error handling is back to what it was before the migration: the `Document` is managed with try-with-resources, and closing it also closes the underlying `PdfDocument`, so the double close and the extra `catch`/`finally` block are gone (@stetsche, @copilot).
- **Rendering failures** — `log.error("Error rendering PDF", e)` keeps the stack trace (@copilot).
- **Version properties / convergence / latest** — `itext.version` is 9.7.0 and `itext.html2pdf.version` is 6.3.3, the current releases, and 9.7.0 is exactly what pdfHTML 6.3.3 resolves to, so `kernel`, `layout`, `io`, `forms`, `svg`, `pdfa` and `pdfua` all converge on one version — `mvn dependency:tree` shows no divergence, so a BOM is not needed. Properties are back in the sorted block (@copilot, @stetsche).
- **Title/description** — updated to describe what the PR actually does (@copilot).
- **Regression test** — `NegotiationPdfGenerationTest` now attaches a real docx and asserts the merged export has more pages than the summary-only export, which is the failure you hit on master (@stetsche).

Verified locally on JDK 21: full backend suite green (1080 tests, 0 failures).

Two things I could not settle:

1. **"PROJECT" header page break** — I could not reproduce it. Rendering `negotiation-1` from the integration test data puts `PROJECT`, `TITLE` and `DESCRIPTION` together on page 2, and in synthetic renders sweeping the content height in single-line steps (with sections both shorter and taller than a page) iText honoured `page-break-after: avoid` on the `h2` in every case. Could you share the payload/negotiation you exported? Otherwise I can wrap each heading plus its `.section` in a keep-together container, but I would rather not add that blind.
2. **Squash message** — commit 770a50dc carries a `BREAKING CHANGE:` footer from when it was written; nothing public changed, so it is worth dropping from the squash message so it does not bump the release as a breaking change.

Unrelated to this PR, but visible now that iText logs it: `PDF_NEGOTIATION_SUMMARY.html` has an empty `url('data:image/svg+xml;base64,')` placeholder in the `@top-right` rule, which iText reports as an ERROR on every render (Flying Saucer ignored it silently). Happy to remove that dead CSS here or in a follow-up.

---

CI is green except `codecov/patch` (61.5% on a 13-line patch): the uncovered lines are the two log-only branches — the "no paragraphs found in DOC" warning and the `renderPdf` failure handler — which cannot be reached without a DOC that has zero paragraphs, respectively without mocking iText itself. `codecov/project` reports -0.07%, measured from a single upload session against a base with four, so it is noise rather than a real drop. `DocConverterTest` was added on top: `DocConverter` was previously only reached through the conversion service with invalid bytes, so the DOC to PDF path itself had no test at all — it now converts the existing `test-valid.doc` fixture and asserts the resulting PDF carries text.

stetsche (approved):
@fuchs-an Code looks good now, thanks for your work on this! You did some basic testing locally, right? I'd do some additional testing once merged together with #1250 .

fuchs-an:
@stetsche Yes, locally I did some basic testing. From my POV everything works as expected now.

stetsche:
Resolved the conflicts on pom.

PR #1234, opened 2026-07-13
feat: add collaborators as requestors

ChrisiSailer:
### Description:

Users can now be added as collaborators to a negotiation, allowing them to edit the negotiation and invite additional collaborators. When a negotiation is transferred to a collaborator, the existing collaborator access link is automatically terminated. Only the negotiation author has permission to remove other collaborators.

### Checklist:

_Make sure you tick all the boxes below if they are true or do not apply before you ask for review_

**Required for all pull requests:**
- [x] I have performed a self-review of my code
- [x] I have made my code as simple as possible
- [x] I have removed all commented code
- [x] I have described the PR and added a meaningful title in the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
**If applicable to this PR:**
- [x] I have added relevant tests for my changes and the code coverage has not dropped substantially
- [x] I have updated the documentation in all relevant places (Javadoc, Swagger, MDs...)

PR #1237, opened 2026-07-17
fix: added authorization for admins to remove resources from negotiation

### LINKED ISSUE #1219: [BUG] Admin cannot remove resources from a draft negotiation they did not create
**Describe the bug**
An admin who is not the author of a negotiation is blocked from removing resources when the negotiation is in DRAFT state. The service-layer precondition check in `verifyRemoveResourcePreconditions` only allows the negotiation creator, with no admin bypass. This is inconsistent with every other similar check in `NegotiationServiceImpl` — admins can add resources to the same draft (PATCH endpoint) and can delete the entire draft negotiation, but cannot remove a single resource from it.

Root cause: `NegotiationServiceImpl.verifyRemoveResourcePreconditions` (introduced in #992) checks only `isNegotiationCreator`, whereas `deleteNegotiation` and `addAnyNewResourcesToNegotiation` both include `|| isCurrentlyAuthenticatedUserAdmin()`.

**How to reproduce**
1. As a researcher, start a new negotiation with multiple resources but do not submit it (leave it in DRAFT state).
2. As an admin (who is not the negotiation author), open the negotiation and click the edit button.
3. In the resources section, click the delete (trash can) button on one of the resources.

**What happens**
The deletion fails with an error message. The backend returns 403 Forbidden.

**What should happen**
The resource is removed successfully. Admins should be able to remove resources from draft negotiations, consistent with their ability to add resources and delete the draft entirely.

**Acceptance criteria**
- An admin can remove a resource from a negotiation that is in DRAFT state, regardless of whether they are the negotiation author.
- The negotiation author can still remove resources from their own draft negotiation (existing behaviour is preserved).
- Removing the last remaining resource is still rejected with an appropriate error for both admin and author.
- The fix aligns `verifyRemoveResourcePreconditions` with the pattern used in `deleteNegotiation` (i.e. `!isNegotiationCreator && !isAdmin` → throw).

ChrisiSailer:
### Description:

Admins now have the proper authorization to remove resources from a negotiation which they do not own. 

### Checklist:

**If applicable to this PR:**

### DIFF
--- a/backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationServiceImpl.java
+++ b/backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationServiceImpl.java
@@ -386,9 +386,10 @@ public class NegotiationServiceImpl implements NegotiationService {
   }
 
   private void verifyRemoveResourcePreconditions(String negotiationId, Negotiation negotiation) {
-    if (!isNegotiationCreator(negotiationId)) {
+    if (!isNegotiationCreator(negotiationId)
+        && !AuthenticatedUserContext.isCurrentlyAuthenticatedUserAdmin()) {
       throw new ForbiddenRequestException(
-          "Only the negotiation author can remove resources from a draft negotiation");
+          "Only the negotiation author and admins can remove resources from a draft negotiation");
     }
     if (negotiation.getCurrentState() != NegotiationState.DRAFT) {
       throw new IllegalStateException(
--- a/backend/src/test/java/eu/bbmri_eric/negotiator/negotiation/NegotiationControllerTests.java
+++ b/backend/src/test/java/eu/bbmri_eric/negotiator/negotiation/NegotiationControllerTests.java
@@ -1636,9 +1636,58 @@ public class NegotiationControllerTests {
   }
 
   @Test
+  @WithUserDetails("admin")
+  @Transactional
+  void removeResource_draftStatus_multipleResources_adminCanRemoveResource() throws Exception {
+    // negotiation-5 belongs to TheResearcher and has 2 resources (ids 5 and 7)
+    Negotiation negotiation = negotiationRepository.findById(NEGOTIATION_5_ID).get();
+    negotiation.setCurrentState(NegotiationState.DRAFT);
+    negotiationRepository.saveAndFlush(negotiation);
+
+    int initialResourceCount = negotiation.getResources().size();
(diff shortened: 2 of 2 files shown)


backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationServiceImpl.java
```diff
@@ -386,9 +386,10 @@ public void removeResourceFromNegotiation(String negotiationId, Long resourceId)
   }
 
   private void verifyRemoveResourcePreconditions(String negotiationId, Negotiation negotiation) {
-    if (!isNegotiationCreator(negotiationId)) {
+    if (!isNegotiationCreator(negotiationId)
```
stetsche, line 389: We might need to queue the PR, but would be nice to use `isNegotiationEditor` here as well. 1) to have it defined in one place and 2) so collaborators can remove resources.

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationServiceImpl.java
```diff
@@ -386,9 +386,10 @@ public void removeResourceFromNegotiation(String negotiationId, Long resourceId)
...
+        && !AuthenticatedUserContext.isCurrentlyAuthenticatedUserAdmin()) {
       throw new ForbiddenRequestException(
-          "Only the negotiation author can remove resources from a draft negotiation");
+          "Only the negotiation author and admins can remove resources from a draft negotiation");
```
stetsche, line 392: 1) simplify to editors
2) this applies to all negotiations, not just drafts, is that correct?

```suggestion
          "Only the negotiation editors can remove resources from a negotiation");
```

stetsche (changes requested):
Looks good, but would be better to incorporate https://github.com/BBMRI-ERIC/negotiator/pull/1234

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationServiceImpl.java
ChrisiSailer: Collaborators can’t remove resources from a resource, so the “editor” role wouldn’t be appropriate. As far as I know, resource removal should only be allowed in the draft state.

konradlang (approved):

stetsche (approved):

PR #1243, opened 2026-07-30
fix: slow responsiveness for large negotiations

### LINKED ISSUE #808: [BUG] Slow reponsivness for large Negotiations
**Describe the bug**
If you have a negotiation with 1000s of resources, the UI is very slow. Should be fixed

ChrisiSailer:
### Description:

Added pagination to resource loading on the negotiation page to improve performance with large datasets. Refactored resource fetching to load resources per organization instead of at the negotiation level for better scalability and structure.

### Checklist:

**If applicable to this PR:**

### DIFF
--- a/backend/src/main/java/eu/bbmri_eric/negotiator/governance/organization/OrganizationForNegotiationDTO.java
+++ b/backend/src/main/java/eu/bbmri_eric/negotiator/governance/organization/OrganizationForNegotiationDTO.java
@@ -2,34 +2,17 @@ package eu.bbmri_eric.negotiator.governance.organization;
 
 import eu.bbmri_eric.negotiator.negotiation.state_machine.resource.NegotiationResourceState;
 import io.swagger.v3.oas.annotations.media.Schema;
-import java.util.Objects;
-import lombok.AllArgsConstructor;
-import lombok.Getter;
-import lombok.NoArgsConstructor;
-import lombok.Setter;
+
+import lombok.*;
 import org.springframework.hateoas.server.core.Relation;
 
-@Getter
-@Setter
+@Data
+@EqualsAndHashCode(callSuper = true)
 @AllArgsConstructor
 @NoArgsConstructor
 @Relation(collectionRelation = "organizations", itemRelation = "organization")
 @Schema(description = "An organization")
 public class OrganizationForNegotiationDTO extends OrganizationDTO {
   boolean updatable;
-
   NegotiationResourceState status;
-
-  @Override
-  public boolean equals(Object o) {
-    if (o == null || getClass() != o.getClass()) return false;
-    if (!super.equals(o)) return false;
-    OrganizationForNegotiationDTO that = (OrganizationForNegotiationDTO) o;
-    return updatable == that.updatable && status == that.status;
-  }
-
-  @Override
-  public int hashCode() {
-    return Objects.hash(super.hashCode(), updatable, status);
-  }
 }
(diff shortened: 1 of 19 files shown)


backend/src/main/java/eu/bbmri_eric/negotiator/governance/organization/OrganizationForNegotiationDTO.java
```diff
@@ -0,0 +1,35 @@
...
+import eu.bbmri_eric.negotiator.negotiation.state_machine.resource.NegotiationResourceState;
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.util.Objects;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import org.springframework.hateoas.server.core.Relation;
+
+@Getter
```
stetsche: use lombok [`@Data`](https://projectlombok.org/features/Data), remove `@Getter`, `@Setter`, `#equals`, `#hashCode`.

frontend/src/views/NegotiationPage.vue
```diff
@@ -245,21 +245,21 @@
...
+                    :resources-last-updated="resourcesLastUpdated"
+                    @reload-resource-info="reloadResources()"
                   />
                 </div>
               </div>
             </div>
           </li>
         </ul>
         <NegotiationPosts
+          data-cy="negotiation-posts"
```
stetsche: I'd like to use `data-test` to keep it generic, rather than "cypress branded". We might switch in the future, or add component scoped unit tests, using vitest for example.

backend/src/main/java/eu/bbmri_eric/negotiator/governance/organization/OrganizationRepository.java
```diff
@@ -34,4 +35,32 @@ public interface OrganizationRepository
...
+                          WHEN 'RETURNED_FOR_RESUBMISSION' THEN 3
+                          WHEN 'CHECKING_AVAILABILITY' THEN 4
+                          WHEN 'RESOURCE_UNAVAILABLE_WILLING_TO_COLLECT' THEN 5
+                          WHEN 'RESOURCE_UNAVAILABLE' THEN 6
+                          WHEN 'RESOURCE_AVAILABLE' THEN 7
+                          WHEN 'ACCESS_CONDITIONS_INDICATED' THEN 8
+                          WHEN 'ACCESS_CONDITIONS_MET' THEN 9
+                          WHEN 'RESOURCE_NOT_MADE_AVAILABLE' THEN 10
+                          WHEN 'RESOURCE_MADE_AVAILABLE' THEN 11
+                      END DESC
```
stetsche, line 61: How is that org-state computed currently? I would like to avoid hardcoding those events here.

stetsche (changes requested):

backend/src/main/java/eu/bbmri_eric/negotiator/governance/organization/OrganizationRepository.java
ChrisiSailer: Should this be fixed with configurable state machine in future?

stetsche: What functionality would we loose if we keep this logic on the frontend? Sorting by state? On which screen?

stetsche: While working on the configurable state machine, I noticed that that this "progress order" is defined as the enum's ordinal value, so the position in the enum defines the order, it's not specified on the frontend. I'd suggest a similar query but to do the state progress comparison as java code.

PR #1250, opened 2026-07-31
feat: add feature flag for exports

### LINKED ISSUE #1223: [FEATURE] Feature flag for exports
**Brief description.**
PDF exports are not a core negotiator feature and go against the design of keeping negotiation data within the system. We should add a feature flag for single negotiator instances to opt into the export functionality if required. By default, this functionality would we disabled

- Adds single spring configuration property to enable export functionality
- Export is disabled by default
When disabled:
- Frontend won't show export buttons
- Endpoints won't return exports
When enabled:
- Same behavior as before

ChrisiSailer:
### Description:

Introduced the `negotiator.export.enabled` Spring property to control whether a Negotiator instance allows PDF negotiation downloads.

### Checklist:

**If applicable to this PR:**


stetsche (changes requested):
I like the approach on how the config lands on the frontend, I think this is how we should load ui-config and the other feature flags at some point - with one api request though. For the moment I feel like we should not add an api call for every setting though. Here, we are blocking the app mount, even the login page, before we get a particular setting for an export, which is disabled by default, so I think we would be better off using a static variable from the env at build time - see other feature flags consumed on the front end (also https://github.com/BBMRI-ERIC/negotiator/pull/1232).

Also it would be nice to use the `FEATURE_FLAG_` env variable pattern and document it. However, for the backend, i think we can keep the current NEGOTIATOR_EXPORT_ENABLED setting.
The downside is that we will have two variables for the same thing, but I think it's fair, and we can map both to one feature flag variable on deployment. To be documented at:
https://bbmri-eric.github.io/negotiator/deployment#custom-properties-for-the-backend-docker-container
https://bbmri-eric.github.io/negotiator/deployment#custom-properties-for-the-frontend-docker-container

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationController.java
```diff
@@ -95,6 +96,9 @@ public NegotiationController(
     this.negotiationPdfService = negotiationPdfService;
   }
 
+  @Value("${negotiator.feature-flags.pdf-export-enabled:${negotiator.export.enabled:false}}")
```
stetsche:
```suggestion
  @Value("${negotiator.feature-flags.pdf-export-enabled:false}")
```

docs/deployment.md
```diff
@@ -79,21 +79,22 @@ the [application file](https://github.com/BBMRI-ERIC/negotiator/blob/master/back
...
-| `SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_AUDIENCES` | OAuth2 Resource Audience(s); see [aud claim](https://datatracker.ietf.org/doc/html/rfc8693#section-3.1) | "https://negotiator.bbmri-eric.eu,negotiator-api"                            |
+| `NEGOTIATOR_WEBHOOK_SECRET_ENCRYPTION_PREVIOUS_MASTER_KEY` | Previous webhook secret master key used only for key rotation (`rotate-secrets` profile).                                                                                       | `""`                                                                          |
+| `NEGOTIATOR_FEATURE_FLAGS_PDF_EXPORT_ENABLED`        | Feature flag that enables/disables negotiation PDF generation endpoint.                                                                                                         | `false`                                                                       |
+| `SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_AUDIENCES` | OAuth2 Resource Audience(s); see [aud claim](https://datatracker.ietf.org/doc/html/rfc8693#section-3.1)                                                                         | "https://negotiator.bbmri-eric.eu,negotiator-api"                            |
```
stetsche: Could we just touch the one added line here? If you want to fix the entire formatting of the table that's fine too, your call

.github/workflows/CI.yml
```diff
@@ -221,6 +221,8 @@ jobs:
         run: docker load --input /tmp/negotiator-frontend.tar
 
       - name: Setup API server
+        env:
+          FEATURE_FLAG_PDF_EXPORT_ENABLED: true
```
stetsche, line 225: Could we rather set this in `compose.yml`? Then this will pass on a local e2e test run as well.

docs/deployment.md
ChrisiSailer: The formatting fix was applied automatically due to the longer new line. Therefore, I would keep it as is.

.github/workflows/CI.yml
ChrisiSailer: Since the requirement states that the default setting should be false, I decided to explicitly enable it for the E2E test run.

stetsche: Could we skip the test(s) when `FEATURE_FLAG_PDF_EXPORT_ENABLED: false`? Then we avoid e2e test failures when this is not provided, I think that would be the best compromise. Sill keeping `FEATURE_FLAG_PDF_EXPORT_ENABLED: true` for CI though.

stetsche (changes requested):
Looking good, just one request: https://github.com/BBMRI-ERIC/negotiator/pull/1250#discussion_r3879343556

stetsche:
Rebased and addressed my comment. @konradlang , could you review please?

stetsche (approved):

konradlang (approved):

PR #1265, opened 2026-08-14
feat: add helpdesk actor tracking for posts and resource lifecycle transitions

ondraholis:
### Description:

The pull request covers the following changes

  - Adds the `negotiator_helpdesk` scope mapped to `ROLE_HELPDESK_INTEGRATION` authority
    in `CustomJWTAuthConverter`. Registers the scope in the OIDC server config and adds a dedicated
    `helpdesk-client` to the dev/test data.

  - Adds a `helpdesk_actor` column to the `post` table. When a caller holding
    `ROLE_HELPDESK_INTEGRATION` supplies a `helpdeskActor` value in the request body, it is persisted
    on the `Post` entity and returned in `PostDTO`. Regular users cannot set this field. The
    `TimelineItem.vue` component prefers `helpdeskActor` over `createdBy.name` when present.

  - Adds a `helpdesk_actor` column to `negotiation_resource_lifecycle_record`. The value
    is threaded through the state machine message headers down to `Negotiation.buildResourceStateChangeRecord()`
    for both the single-resource (`PUT …/lifecycle/{event}`) and bulk (`PATCH …/resources`) paths.
    `getTriggeredBy()` falls back to `createdBy.name` when `helpdeskActor` is null. Regular callers
    cannot spoof the field — it is stripped unless the caller holds `ROLE_HELPDESK_INTEGRATION`.

  ### Checklist:

  **If applicable to this PR:**


backend/src/main/java/eu/bbmri_eric/negotiator/common/AuthenticatedUserContext.java
```diff
@@ -74,9 +76,22 @@ public void runAsSystemUser(Runnable task) {
...
-            .anyMatch(a -> a.getAuthority().equals("ROLE_ADMIN"));
+    return hasRole("ROLE_ADMIN");
+  }
+
+  /**
+   * Check if the current request is authenticated as the helpdesk integration service account.
+   *
+   * @return true if the caller holds the ROLE_HELPDESK_INTEGRATION authority
+   */
+  public static boolean isHelpdeskIntegration() {
```
RadovanTomik, line 87: why not use the `hasRole()`?

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationController.java
```diff
@@ -238,14 +241,19 @@ private static String getOptionalComment(
   public ResponseEntity<?> sendEventForNegotiationResource(
       @Valid @PathVariable String negotiationId,
       @Valid @PathVariable String resourceId,
-      @Valid @PathVariable("event") NegotiationResourceEvent event) {
+      @Valid @PathVariable("event") NegotiationResourceEvent event,
+      @RequestHeader(value = "X-Helpdesk-Actor", required = false) String helpdeskActor) {
```
RadovanTomik, line 245: why is it a header here and a request body field in `PostCreateDTO`?

RadovanTomik (changes requested):
Hey, thanks for opening the PR! Went over the code and realised that the helpdesk actor field would be distributed throughout a relatively large part of the codebase, violating some best practices. The actor would need to be added to all actions that a representative can do (this applies to all future actions that would be added). If we stuck with the simple "Graz Biobank marked resource X as checking availability", it would remove a lot of lines of code/complexity and still work for us. If a helpdesk user wants their name to be known, they can simply sign their comment. Or if you can suggest a way to scope the changes/make it work so the "actor" is saved in a separated/scoped way, I'm all ears.

stetsche:
I agree with @RadovanTomik 's comment. The post and resource lifecycle domains should not need to worry about an helpdesk actor. To me using just the system account, showing the system user's name in the UI is fine at this stage. One way this could be solved and it might become relevant with more integrations in the future, would be adding the concept of external uses - adding person rows in the db for users of external integrations. I would see it as a possible future want though, not something we need now.

ondraholis:
Thank you for the review guys. After the discussion with Radovan, I am closing this ticket. I will keep it out of scope from the implementation for now.

PR #1267, opened 2026-08-14
feat: implement GET /v3/negotiations/{negotiationId}/posts/{postId} with visibility-aware access control

ondraholis:
### Description:

The GET /v3/posts/{postId} endpoint existed in PostController but was unreachable and non-functional: the path fell through to the catch-all denyAll() security rule, and PostServiceImpl.findById() was an unimplemented stub returning null. This PR implements the endpoint to support the helpdesk integration middleware, which needs to fetch full post text by ID when relaying a negotiation.post.added webhook to the help desk.

The implementation applies the same visibility rules as the existing post-listing endpoint: public posts are readable by any negotiation participant; private posts are only accessible to the negotiation creator, admins, or participants whose organization matches the post's target organization.

### Checklist:

  - [x] I have described the PR and added a meaningful title in the Conventional Commits format

**If applicable to this PR:**

### DIFF
--- a/backend/src/main/java/eu/bbmri_eric/negotiator/common/configuration/security/HTTPRegistryConfigurer.java
+++ b/backend/src/main/java/eu/bbmri_eric/negotiator/common/configuration/security/HTTPRegistryConfigurer.java
@@ -33,8 +33,6 @@ public class HTTPRegistryConfigurer {
         .authenticated()
         .requestMatchers(mvc.pattern("/v3/attachments/**"))
         .authenticated()
-        .requestMatchers(mvc.pattern(HttpMethod.GET, "/v3/posts/**"))
-        .authenticated()
         .requestMatchers(mvc.pattern("/v3/negotiations/**"))
         .authenticated()
         .requestMatchers(mvc.pattern("/v3/access-criteria"))
--- a/backend/src/main/java/eu/bbmri_eric/negotiator/common/configuration/security/HTTPRegistryConfigurer.java
+++ b/backend/src/main/java/eu/bbmri_eric/negotiator/common/configuration/security/HTTPRegistryConfigurer.java
@@ -33,6 +33,8 @@ public class HTTPRegistryConfigurer {
         .authenticated()
         .requestMatchers(mvc.pattern("/v3/attachments/**"))
         .authenticated()
+        .requestMatchers(mvc.pattern(HttpMethod.GET, "/v3/posts/**"))
+        .authenticated()
         .requestMatchers(mvc.pattern("/v3/negotiations/**"))
         .authenticated()
         .requestMatchers(mvc.pattern("/v3/access-criteria"))
--- a/backend/src/main/java/eu/bbmri_eric/negotiator/post/PostServiceImpl.java
+++ b/backend/src/main/java/eu/bbmri_eric/negotiator/post/PostServiceImpl.java
@@ -120,8 +120,11 @@ public class PostServiceImpl implements PostService {
     verifyReadAccess(negotiationId);
     if (!post.isPublic() && !isNegotiationCreatorOrAdmin(negotiationId)) {
       Person user = getCurrentUser();
-      Set<Organization> accessibleOrganizations = getUserAccessibleOrganizations(user);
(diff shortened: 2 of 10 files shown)


backend/src/main/java/eu/bbmri_eric/negotiator/post/PostServiceImpl.java
```diff
@@ -113,8 +113,19 @@ private void checkAuthorization(
   }
 
   @Override
+  @Transactional
   public PostDTO findById(String id) {
```
stetsche: 1. Please move `verifyReadAccess` above `findById`
2. Please check if organization level access check could be implemented like this: Rather than checking `user -> resources -> oragnizations` and `user -> networks -> resources -> oragnizations` against `post -> organization`, do `post -> organization -> users (representatives)` against `user`

stetsche (changes requested):
Good catch, thanks for the PR! One optimization that i think is worth checking out:

backend/src/main/java/eu/bbmri_eric/negotiator/common/configuration/security/HTTPRegistryConfigurer.java
```diff
@@ -33,6 +33,8 @@ public void configure(
         .authenticated()
         .requestMatchers(mvc.pattern("/v3/attachments/**"))
         .authenticated()
+        .requestMatchers(mvc.pattern(HttpMethod.GET, "/v3/posts/**"))
+        .authenticated()
```
stetsche:
```suggestion
```
Could we instead move the endpoint to ` GET /negotiations/{negotiationId}/posts/{postId}`, same as `PUT /negotiations/{negotiationId}/posts/{postId}`, to mirror`?

ondraholis: If I would move the endpoint to `GET /negotiations/{negotiationId}/posts/{postId}`, I would need to add negotiationId (as non-null) into PostDTO since the PostModelAssembler.toModel would need to change its implementation to:
```
@Override
  public @NonNull EntityModel<PostDTO> toModel(@NonNull PostDTO entity) {
    List<Link> links = new ArrayList<>();
    links.add(linkTo(methodOn(PostController.class).getById(entity.getNegotiationId(), entity.getId())).withSelfRel());
    return EntityModel.of(entity, links);
  }
```
and maybe some other changes after that would be needed (atleast in the tests). Are you sure that you want this change and can I proceed with it?

backend/src/main/java/eu/bbmri_eric/negotiator/post/PostServiceImpl.java
ondraholis: Done in commit [ea45e50](https://github.com/BBMRI-ERIC/negotiator/pull/1267/commits/ea45e50250ebd6bbf927a21c77d27f81137c473c)

backend/src/main/java/eu/bbmri_eric/negotiator/common/configuration/security/HTTPRegistryConfigurer.java
stetsche: I think this is fine. A post is always associated with an negotiation. The small refactor buys us consistency and now that we start using this endpoint, it will be harder in the future to change it.

backend/src/main/java/eu/bbmri_eric/negotiator/post/PostServiceImpl.java
stetsche: Nice!

backend/src/main/java/eu/bbmri_eric/negotiator/common/configuration/security/HTTPRegistryConfigurer.java
ondraholis: @stetsche  Done in commit [6562469](https://github.com/BBMRI-ERIC/negotiator/pull/1267/changes/65624694d04bcc7ff8c9cebc18dded8ba684fc1c)

stetsche (approved):
Nice work, thanks for the contribution!

PR #1269, opened 2026-08-15
feat: add involvedOrganizationExternalIds to negotiation.state.updated webhook event

ondraholis:
### Description:

The negotiation.state.updated webhook payload now includes involvedOrganizationExternalIds — a set of external IDs of all organizations whose resources are part of the negotiation.

This field is required by the negotiator-link middleware to resolve the correct OrganizationConfig entry (which holds credentials) before making any secondary Negotiator API call. Without knowing which organization external IDs are involved, the middleware cannot identify which configured organization the webhook belongs to and has no credentials to fetch negotiation data or resources. With the IDs present in the payload, the middleware can immediately resolve the config; if none of the IDs match its YAML configuration, the event is silently discarded without any API call.

  Changes:
  - `NegotiationStateChangeEvent` carries the org external IDs collected from the negotiation's
    resources inside `PersistStateChangeListener.onPersist()`, which runs within a `@Transactional`
    boundary so lazy-loading is safe
  - `NegotiationStateUpdatedWebhookEvent` record extended with the new field; mapping strategy
    threads it through
  - Tests updated to compile with the new constructor parameter; `sendEvent_approveNewNegotiation_isOngoing`
    asserts the published event contains the correct org IDs; `publishNegotiationStateChangeEvent_dispatchesToActiveWebhooks` asserts the field is present in the serialized webhook JSON body

### Checklist:

**If applicable to this PR:**

### DIFF
--- a/backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/state_machine/negotiation/NegotiationStateChangeEvent.java
+++ b/backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/state_machine/negotiation/NegotiationStateChangeEvent.java
@@ -1,5 +1,6 @@
 package eu.bbmri_eric.negotiator.negotiation.state_machine.negotiation;
 
+import java.util.Set;
 import lombok.Getter;
 import org.springframework.context.ApplicationEvent;
 
@@ -10,17 +11,20 @@ public class NegotiationStateChangeEvent extends ApplicationEvent {
   private final NegotiationState fromState;
   private final NegotiationState toState;
   private final NegotiationEvent event;
+  private final Set<String> involvedOrganizationExternalIds;
 
   public NegotiationStateChangeEvent(
       Object source,
       String negotiationId,
       NegotiationState fromState,
       NegotiationState toState,
-      NegotiationEvent event) {
+      NegotiationEvent event,
+      Set<String> involvedOrganizationExternalIds) {
     super(source);
     this.negotiationId = negotiationId;
     this.fromState = fromState;
     this.toState = toState;
     this.event = event;
+    this.involvedOrganizationExternalIds = involvedOrganizationExternalIds;
   }
 }
--- a/backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/state_machine/negotiation/PersistStateChangeListener.java
+++ b/backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/state_machine/negotiation/PersistStateChangeListener.java
@@ -9,7 +9,10 @@ import eu.bbmri_eric.negotiator.user.Person;
 import eu.bbmri_eric.negotiator.user.PersonRepository;
 import jakarta.transaction.Transactional;
 import java.time.LocalDateTime;
+import java.util.Collections;
 import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
 import lombok.extern.apachecommons.CommonsLog;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.lang.Nullable;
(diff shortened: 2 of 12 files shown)


stetsche:
I think this information should rather be queried from the REST API, than added as context for this events. The event provides the negotiationId, which should be enough context. If this information is not provided by the API, we should think of which endpoint would make sense for this. @ondraholis , is there any issue with getting that info from the REST API? @RadovanTomik , what do you think?

RadovanTomik:
hey, I agree with @stetsche . The middleware can pull information from the REST API with the negotiation ID. If you want to scope relevant negotiations, I would opt for linking a webhook to an Organization/list of resources; this way, the webhook would fire only if the negotiation involves particular resources/organizations.

stetsche:
Thanks for the alternative suggestion @RadovanTomik , adding an organization scope on webhook creation/edit would be a nice addition for webhooks. @ondraholis , if want to implement that, rather than doing the filtering on the middleware side, I can help with defining some acceptance criteria for it.

ondraholis:
Hi @RadovanTomik @stetsche  ,

Unfortunately, it is not possible for the middleware to retrieve information from the REST API using only the negotiation ID, for the following reason:

The middleware will require every organisation using the integration to have its own service account in Negotiator, which will be linked to the resources belonging to that organisation. Therefore, the middleware will always need to authenticate and obtain a token using the credentials of the respective organisation. Using this token, it will only be able to access negotiations that contain at least one resource belonging to that organisation.

When the middleware receives a particular negotiation ID for the first time and needs to create a ticket for it in the help desk, however, the middleware does not know which organisation it should use to access the negotiation and retrieve its data.

The alternative that Radovan is proposing is definitely a good solution, but it is somewhat too specific. Ideally, it would be better to extend the webhook functionality as a whole so that it works similarly to how webhooks work in most help desk systems. Here is an example of how it works in Zammad, where you can define exactly under which conditions a particular webhook should be sent:

image

In addition, you can create a custom payload for the webhook and specify exactly which information you want to include in it.

If something similar were introduced in Negotiator, I could define a webhook for each organisation using the help desk integration, for example:

`POST /webhooks/negotiator/{organizationExternalId}`

This webhook could then be triggered whenever a negotiation is changed (or, more specifically, when certain types of changes occur).

However, implementing this kind of webhook functionality would require a significantly larger development effort. On the other hand, if only the functionality proposed by @RadovanTomik were implemented, it would be unnecessarily specific and would still require a relatively large number of changes (including UI) considering that it would only address my particular use case.

Therefore, I would only like to mention this webhook extension as a possible future improvement in my master's thesis and not implement it as part of my work.

Another way to solve the whole issue would be to introduce a new endpoint:

`GET /negotiations/{id}/organizations`

This endpoint could return a Set of involved organizationExternalIds for the given negotiation ID, with the only requirement being that the user is authenticated in order to receive the result.

This approach has the disadvantage that it somewhat violates confidentiality, as it would allow authenticated users to find out which organisations are involved in a particular negotiation. On the other hand, this is not particularly sensitive information, so I think it could be acceptable.

What do you think about this overall? Could I proceed with the existing PR, or should I create a PR for the new endpoint to unblock my integration? I would really prefer not to implement the additional webhook functionality at this point.

Thanks for your opinion.

stetsche:
Hi @ondraholis ,
I see now why getting the involved organizations from the API is an issue, as the service users are organization scoped. One option, which is not ideal for sure, would be to try something like `GET /negotiations/{id}` and handle the `403`. This would need to be done with the service user of each organization handled by the middleware, which makes it even less ideal, but the number will probably be 1 initially.

I would like to avoid exposing the involved organizations to users that don't have access to the negotiation.

As for the option of extending webhooks, which would be cleaner, I see it as pretty valuable and not to specific to this integration, as probably most webhook consumers will actually be organization scoped. We actually discussed this feature before implementing webhooks, but opted to keep it minimal initially and extend when needed. This confirms now, that this is a feature we will need. But I don't think we currently want to develop and maintain something as complex as Zammad's conditions for webhooks and rather keep it simple. One scope dropdown for organization. Maybe later, if needed, we can scope by a list of events as well.
I see how this could be too big of an effort and out of scope for your work. Possibly I can pick this up while you are still working on this integration though, I am not sure at the moment though.

ondraholis:
Hi @stetsche,

Thank you for your answer.

Since I am still at the beginning and need to first implement an MVP, I will use the approach of querying GET /negotiations/{id} with each service user from the organization. The webhook improvement can be kept as a future enhancement. Maybe you, I, or someone else will have the capacity to implement it in the future. After that, it will not be a problem to adopt it in the middleware and use it instead of querying with each service account from the organization.

Therefore, I believe I can close this PR. Thank you for your time.

PR #1286, opened 2026-09-04
chore: yarn version 1 to 4

### LINKED ISSUE #1190: [FEATURE] upgrade yarn 1 to 4
**Brief description.**

- Faster Vite Boot Times: Vite relies on fast dependency resolution. Yarn 4's Plug'n'Play (PnP) maps dependencies directly, allowing Vite to resolve and load your Vue components faster during local development.

- Instant CI/CD Deploys: Vue projects often pull in large dependency trees (Vite, dynamic linters, Component libraries). Yarn 4's "Zero-Installs" lets you check your cache into Git, reducing deployment times on platforms like Vercel, Netlify, or GitHub Actions to nearly zero seconds.

- Safer Vue Ecosystem Upgrades: The Vue ecosystem relies heavily on peer dependencies (e.g., matching vue, @vue/compiler-sfu, and component libraries). Yarn 4 enforces strict peer dependency checks, preventing runtime crashes caused by silent version mismatches.

- Cleaner Monorepo Management: If you separate your Vue frontend, Nuxt SSR engine, and backend into a monorepo, Yarn 4 Workspaces manage cross-project dependencies seamlessly without symlink bugs.

- No node_modules Corruption: Heavy Vue UI frameworks (like Vuetify or PrimeVue) can sometimes break during sequential yarn install updates in Yarn 1. Yarn 4 uses immutable .zip archives, ensuring your local environment never gets corrupted.

Replace all yarn 1 with 4, including Docker and CI

### LINKED ISSUE #1285: [FEATURE] Yarn Version 1 to Version 4 Migration
# Plan: Pin Yarn Berry (v4) for frontend/ properly

## Context

`frontend/` already has a half-finished local Yarn Berry setup: `.yarnrc.yml` + a vendored `.yarn/releases/yarn-1.22.22.cjs` binary exist on disk, but a blanket `.gitignore` rule (root `.gitignore:51-55`) has silently excluded both from git the entire time. So this pin has never reached CI, Docker, or any other contributor's machine — every environment instead resolves `yarn` to whatever global Yarn happens to be preinstalled (GitHub's hosted runner images and the `node:lts-krypton` Docker base image both ship a global classic Yarn on PATH; one contributor machine checked already defaults to Berry 4.17.1 globally instead — this varies contributor to contributor). The goal: durably commit a real Yarn 4.x (Berry) pin for `frontend/` so local dev, CI, and Docker all resolve to the exact same, version-controlled binary, and confirm nothing downstream breaks.

Verified against current repo state (branch `master`, clean working tree): the `.gitignore` block, the stray `yarn` dependency, the CI job names/cache config, and the Dockerfile all check out as described below. Two corrections to keep in mind throughout:

1. **Node version: CI is exact-pinned, Docker floats — don't conflate the two.** CI's `setup-node` pins the exact patch `24.11.0` in all 5 jobs that use it (confirmed `node-version: '24.11.0'`). The Dockerfile's `FROM node:lts-krypton` is a codename-based LTS tag that floats to whatever the latest Node 24.x patch happens to be at build time — nothing confirms it resolves to the same exact patch CI uses. Both are on the Node 24 line, so there's no practical Node-version risk to this migration, but state it precisely: CI is pinned to `24.11.0`; Docker floats on Node 24 via `node:lts-krypton`, not identically pinned.
2. **CI's E2E job (`test`) is one job definition, but matrix-expanded into 2 parallel runs per trigger** — `browser: [chrome, electron]`, not chrome/firefox. `cy:run:firefox` exists in `package.json` but is never invoked by CI, only available locally. This affects local regression (Step 7 below): both `cy:run` and `cy:run:chrome` need to run, not either/or.

**Scope:** `frontend/` plus the root-level config that supports it (root `.gitignore`, `.github/workflows/CI.yml`) — `docs/` (VitePress, builds via npm/`package-lock.json` in CI) and `backend/` are untouched by this migration.

---

## Step 0 — Branch hygiene

Don't trust any snapshot of "current branch state" baked into a plan — check live, immediately before starting:

```
git checkout master && git pull
git status            # confirm clean
git checkout -b <migration-branch-name>
```

No stash or drop should be needed under normal circumstances.

## Step 1 — Remove stray cruft from frontend/package.json

Delete `"yarn": "^1.22.19"` from `dependencies` (present just before `devDependencies`). Yarn is tooling, not app code — it should never be listed as a project dependency.

## Step 2 — Fix root .gitignore (lines 51-55) + protect the vendored binary

Current block:

```
### Yarn ###
frontend/.yarn
frontend/.yarnrc.yml
# Currently not checked in
frontend/yarn.lock
```

This is why the local pin has never reached git (the `# Currently not checked in` comment is itself already stale — `frontend/yarn.lock` is in fact tracked). Replace with the standard Berry allowlist:

```
### Yarn ###
.yarn/*
!.yarn/releases
!.yarn/plugins
!.yarn/patches
!.yarn/sdks
.yarn/cache
.yarn/unplugged
.yarn/install-state.gz
.pnp.*
```

This is the load-bearing step — without it the pin gets silently dropped again on the next commit.

While in here, add a `.gitattributes` entry so the ~5MB minified binary you're about to commit doesn't get line-ending-mangled or pollute diff/language stats:

```
frontend/.yarn/releases/*.cjs binary linguist-vendored
```

## Step 3 — Upgrade the pin

```
cd frontend
yarn set version stable   # or a specific 4.x
```

The vendored classic 1.22.22 binary forwards this command to Berry's release manager directly — no Corepack needed for this step. If it errors, fall back once to `corepack enable && corepack prepare yarn@stable --activate`, then retry — Corepack is only a one-time bootstrap; once `yarnPath` is rewritten, every later invocation (local, CI, Docker) resolves through the committed pin.

This rewrites `.yarnrc.yml`'s `yarnPath` and drops a new binary in `.yarn/releases/` (replacing `yarn-1.22.22.cjs`) — `nodeLinker: node-modules`, `approvedGitRepositories`, `enableScripts`, `npmMinimalAgeGate` are untouched. It's a version bump, not a switch to PnP.

**It also automatically writes `"packageManager": "yarn@<resolved-version>"` into `package.json`** as a side effect — don't hand-add it. Just verify it landed correctly:

```
grep packageManager frontend/package.json
```

Optional, while in there: add `"engines": { "node": ">=20" }` to `package.json`. Purely a local-dev guardrail — CI and Docker are already on Node 24, this just fails loudly for a contributor on an old Node.

## Step 4 — Regenerate the lockfile

```
yarn install
```

Converts `frontend/yarn.lock` from the classic v1 format to Berry's — expect a full rewrite (large diff). Commit as-is, don't hand-edit. Skim the diff for any dependency whose *resolved version* actually changed (not just reformatted) — Berry's resolver isn't bit-for-bit identical to Classic's.

This exact rewrite is what makes any currently-open Dependabot PR against `frontend/yarn.lock` conflict or go stale — see the pre-merge checklist (Step 11).

## Step 5 — Peer-dependency / hoisting triage (the real risk)

Berry's node-modules linker hoists more strictly than Yarn 1. `frontend/README.md` (lines 3-5) already documents, independently, that this dependency tree has known peer-dependency friction — treat that as corroboration, not just a hypothesis. Watch for new peer-dependency warnings/errors and anything relying on Yarn 1's looser "phantom dependency" hoisting. Given the dependency set, expect the most likely friction from: `vue3-form-wizard` (pinned exact 1.1.1), `vuedraggable` (pinned exact 4.1.0), the `@codemirror/lang-vue` / `codemirror` / `vue-codemirror` trio, and `vue-router`. Triage each warning individually rather than mass-adding dependencies to work around hoisting.

## Step 6 — Update frontend/README.md

Add 1-2 lines noting the pin and that Corepack isn't required, e.g.:

> Yarn is pinned via `.yarnrc.yml` / `.yarn/releases` to version X.Y.Z (Berry). This resolves automatically on `yarn install`; Corepack is not required.

Low cost, high value: the README is the first place a new contributor looks, currently explains *why* Yarn was chosen, but says nothing about the pinned binary they're about to see checked in under `frontend/.yarn/releases/`.

## Step 7 — Full local regression pass

```
yarn build
yarn lint            # eslint . --fix   — mutates files
yarn format          # prettier --write src/ — mutates files
yarn cy:run          # Electron leg — mirrors CI's "electron" matrix instance
yarn cy:run:chrome   # Chrome leg   — mirrors CI's "chrome" matrix instance
```

Run **both** Cypress commands, not either/or — they together are what mirror CI's two matrix legs (see Context above).

`lint` and `format` write to disk rather than only report. After running them, check:

```
git status && git diff --stat -- frontend/src
```

If they produced changes, decide deliberately whether to fold them into this commit (and note it in the PR description) rather than letting unrelated formatting drift silently ride along in a migration diff that should stay reviewable as "Yarn only."

## Step 8 — CI (.github/workflows/CI.yml)

`cache: 'yarn'` + `cache-dependency-path: 'frontend/yarn.lock'` appears in exactly 4 jobs, all via already-SHA-pinned `actions/setup-node` (no change needed there):

- `install` (line 54): `yarn --cwd frontend install`
- `build` (line 73): `yarn --cwd frontend install` / `build`
- `lint-js` (line 110): `yarn install --cwd frontend` — reversed flag order vs. the other three; same effect, worth normalizing while you're in here
- `test` (line 203): `yarn --cwd frontend install` — exists only to give the `cypress-io/github-action` wrapper a local `node_modules` to invoke Cypress from on the bare runner; it does **not** build. The app under test was already built earlier, inside the Docker image, by `build-frontend-image`.

(A 5th `setup-node`, in `build-pages`, is for `docs/` with npm caching — unrelated, don't touch.)

Once `.yarn/releases/*.cjs` + `.yarnrc.yml` are committed, these steps should keep working unchanged — no Corepack step needed. Once you have a real CI run to look at (Step 10), confirm `setup-node`'s built-in yarn cache still resolves correctly against a Berry-format lockfile; if it doesn't, drop the built-in cache and cache `frontend/.yarn/cache` manually via `actions/cache`, or add an explicit `corepack enable` step as a fallback.

Add `--immutable` to all 4 install commands above, so a drifted lockfile fails loudly in CI instead of silently rewriting. This is applied now, in this pre-PR edit pass — it does not wait for a CI run to confirm it first (contrast with the Docker-side `--immutable` in Step 9c, which is deliberately deferred; see there for why).

Add one cheap guard against this exact bug class recurring (a pin silently going untracked again, or an environment silently falling back to a different global Yarn) as an early step in the `install` job:

```
git ls-files --error-unmatch frontend/.yarnrc.yml frontend/.yarn/releases/*.cjs
```

followed by confirming `yarn --cwd frontend --version` matches the version recorded in `package.json`'s `packageManager` field. Near-zero ongoing cost, catches the failure mode this whole migration exists to fix.

Informational, no action needed: `test`'s matrix fan-out means its install/cache-restore runs twice per trigger (once per browser leg); a "cache already exists, skipping save" message from the second leg is benign, not a failure.

## Step 9 — Docker (frontend/Dockerfile + frontend/.dockerignore)

Current: `FROM node:lts-krypton AS build-stage` → `WORKDIR /app` → `COPY . .` → `RUN yarn install` → `RUN yarn build`, then a second `nginxinc/nginx-unprivileged` stage. No Corepack; relies on the same PATH-delegation mechanism as CI. `COPY . .` will now bring in the committed `.yarnrc.yml` + `.yarn/releases` binary automatically, so `yarn install`/`yarn build` should resolve to the pinned Berry version with no other change required for that specific claim — but three concrete improvements belong in this same change:

**9a. Required — skip the wasted Cypress binary download.** `RUN yarn install` has no `--production` flag and `enableScripts: true` is set, so it currently installs every devDependency — including Cypress, triggering a 100+MB postinstall binary download that's never used in this stage (only `vite build` runs here). Because `build-push-frontend` builds both `linux/amd64` and a QEMU-emulated `linux/arm64`, every full release build pays this cost twice, wastefully. Fix it narrowly — `vite` and `sass` are also devDependencies and *are* needed by the build, so a blanket `--production` install would break it:

```dockerfile
ENV CYPRESS_INSTALL_BINARY=0
RUN yarn install
```

Scope this to the Dockerfile build stage only — do **not** set it in CI.yml or globally; the `test` job's separate bare-runner install (Step 8) needs a real Cypress binary to run E2E.

**Framing note for the PR description/commit message:** describe this as a CI-time / build-cost optimization (faster, cheaper `linux/amd64`+`linux/arm64` image builds) — not as a security or shipped-image-size fix. `node_modules` from `build-stage` never reaches the shipped `nginx-unprivileged` stage — only the built `dist/` output is copied over — so this change has zero effect on the final image's size or attack surface. Framing it as a security/size improvement would let a reviewer evaluate it on the wrong axis.

**9b. Layer-cache the install step.** Split:

```dockerfile
COPY package.json yarn.lock .yarnrc.yml ./
COPY .yarn ./.yarn
RUN yarn install
COPY . .
RUN yarn build
```

This lets a source-only commit skip the (now cheaper, post-9a) install layer under `build-push-frontend`'s existing `cache-from/cache-to: type=gha,mode=max` config. Note: `build-frontend-image` (the job that runs on every PR) has **no** `cache-from`/`cache-to` configured at all, so this split doesn't speed up PR builds — its payoff is specifically for `build-push-frontend`'s post-merge/release builds.

Also add to `frontend/.dockerignore` (currently just `.github`, `tests`, `docs`):

```
node_modules
.yarn/cache
.yarn/unplugged
```

This is local-developer hygiene, not a CI fix — root `.gitignore` already keeps `node_modules`/`dist` out of what `checkout` produces, so CI build context and build time are unaffected either way.

**Guardrail: do NOT add `.git` or a broad `.*` glob** to `.dockerignore` — `build-push-frontend` deliberately copies `.git` into the build context (`cp -r .git frontend`) so `build-push-action` can auto-populate OCI revision labels; excluding it would silently stop those labels from being set, with no CI failure to catch it.

**9c. Add `--immutable` to `RUN yarn install`** here too, once 9a/9b are confirmed green — do this last, so any lockfile-drift failure surfaces against an already-stable install step, not tangled up with the restructuring. Concretely, this means a **second, follow-up commit pushed to the same PR after Step 10's initial CI run goes green** — it is deliberately *not* folded into the same pre-PR edit pass that adds `--immutable` to the 4 CI install commands (Step 8). Those two `--immutable` additions are staggered on purpose, not simultaneous. Sequence:

1. Open the draft PR (Step 10) with 9a + 9b already in place, plus Step 8's CI-side `--immutable`. It does **not** yet include this Dockerfile `--immutable` line.
2. Once `build-frontend-image` and the other jobs are confirmed green on that initial push, push a **follow-up commit** to the same PR adding just this one Dockerfile `--immutable` line.
3. Confirm `build-frontend-image` goes green again on that follow-up commit specifically, before moving on to the pre-merge checklist (Step 11).

**Which CI job actually proves this works:** `build-frontend-image` (context: `frontend`, a plain single-arch `docker build` with no yarn/setup-node steps of its own — the whole Berry resolution happens inside Docker) runs automatically on every PR as a dependency of both `test` and `image-scan`. That's your automatic, per-PR proof that `COPY . . + RUN yarn install + RUN yarn build` resolves the pinned binary correctly inside Docker — no separate manual step is strictly required for that claim, though a local `docker build frontend` before opening the PR is still good practice.

**What does NOT prove this works:** a green bare-runner `build` job (Step 8's `yarn --cwd frontend install` / `build`, running directly on the GitHub-hosted runner, outside Docker entirely) is **not** evidence the Docker+Berry path works. It shares no artifact, no image layer, and no filesystem with `build-frontend-image`, and resolves `yarn` in a completely different context — host Node + host-resolved yarn, versus the Dockerfile's own `COPY . .` + `RUN yarn install` inside the build. The two jobs going green are independent signals. Don't let a green `build` check substitute for specifically checking `build-frontend-image` — that's the only job that actually exercises the Docker+Berry path this migration is about.

Separately, accept this as a known gap: `build-push-frontend` — the only job that does the `linux/amd64`+`linux/arm64` QEMU multi-arch build that 9a specifically targets — needs several other jobs and only runs post-merge/on tags, never on a PR. Either run a one-time local sanity check before merging:

```
docker buildx build --platform linux/arm64 -f frontend/Dockerfile frontend
```

or explicitly accept the residual risk and watch the first post-merge build closely (Step 12).

## Step 10 — Push the branch, open a draft PR

Push and open the PR as a draft now — this is what actually produces the CI run Step 8 wants to verify against, and the Docker-in-CI proof Step 9 wants — so do this right after local regression (Step 7) rather than after every other change is "done." At this point the PR carries 9a/9b (Cypress skip + COPY split) and Step 8's CI-side `--immutable`, but **not yet** 9c's Docker-side `--immutable` — that addition is deliberately held back for a follow-up commit, per Step 9c.

Watch: `install`, `build`, `lint-js`, `test` (both chrome and electron legs), `build-frontend-image`, `image-scan`. Iterate here on anything CI-only surfaces (cache resolution, the CI-side `--immutable` failures from Step 8, hoisting issues Electron/Chrome catch that local testing didn't). Remember `build-frontend-image` going green — not `build` — is what proves the Docker+Berry path itself (see Step 9).

Once everything above is green, push the follow-up commit described in Step 9c (Docker-side `--immutable`) and confirm `build-frontend-image` goes green again specifically on that commit before moving to Step 11.

## Step 11 — Pre-merge checklist

- **Dependabot:** check the repo's open PRs for anything from `dependabot[bot]` touching `frontend/yarn.lock` or `frontend/package.json` (the `npm` ecosystem runs monthly against `/frontend`, reviewer `tmilost`, and has recently merged real bumps against the current classic lockfile). Close any that are open — Dependabot recreates them against the new Berry-format lockfile on its next scheduled run, or nudge it immediately with an `@dependabot recreate` comment. Treat this as a live check when you execute the plan, not something already confirmed.
- **Reviewer:** mark the PR ready and explicitly @-request someone with frontend context for the Step 5 hoisting risk — CODEOWNERS won't do this for you. `frontend/` is owned solely by `@tmilost` (the author), so GitHub auto-requests nobody for the frontend-specific hunks; only the root `.gitignore`/`CI.yml` hunks get an automatic reviewer (`@konradlang`), who has no particular frontend stake. Recent history shows Markus Stetschnig and Chrisi as heavy, active frontend/CI/Docker contributors who aren't CODEOWNERS today — request one of them directly. Longer-term, consider adding a second name to CODEOWNERS' `frontend/` line.
- **Team heads-up:** post a short note to active frontend contributors (Markus Stetschnig, Chrisi, RadovanTomik) before merging, covering:
  1. After pulling `master` past this merge, `yarn.lock` will look fully rewritten/conflicting on any branch based on the old lockfile — resolve by taking master's version and re-running `yarn install`, don't hand-merge.
  2. Delete `node_modules` and re-run `yarn install` once after pulling — the resolved Yarn binary is changing, and which global Yarn a machine was falling back to varies.
  3. If `git` refuses a pull/checkout with "untracked working tree file would be overwritten," it means a local, git-invisible `frontend/.yarnrc.yml` or `frontend/.yarn/releases/*.cjs` already exists on that machine predating this migration — remove or rename it and retry.

  Consider timing the merge for a quiet window given how many people actively touch this path.

## Step 12 — Merge and follow up

Merge as a single squash commit so the whole migration (`.gitignore`, lockfile format, `.yarnrc.yml`, `package.json`, Dockerfile, CI config, README) is one revertible unit.

If CI on `master` (push-triggered, plus the daily cron) or `build-push-frontend` fails post-merge: `git revert <merge-sha>` immediately, and hold any pending release/tag until master CI is green again.

Post-merge, specifically:
- Watch the first `build-push-frontend` run (multi-arch, QEMU-emulated arm64) closely — this is the first real exercise of that path, since PR-time CI never runs it.
- Confirm Dependabot's `npm` ecosystem successfully opens a PR against the new Berry-format lockfile on its next scheduled run — its first exposure to a Berry lockfile in this repo.

tmilost:
### Description:

Migrates frontend/ from a half-committed Yarn Berry setup to a properly pinned one. Root .gitignore was silently excluding .yarnrc.yml and .yarn/releases the whole time, so local dev, CI, and Docker were each falling back to whatever global Yarn happened to be installed instead of a pinned version.

Pins Yarn Berry 4.18.0, regenerates the lockfile, hardens CI (--immutable + a guard against this regressing again), and skips a wasted Cypress binary download in the Docker build. 

### Checklist:

**If applicable to this PR:**


frontend/.yarnrc.yml
```diff
@@ -0,0 +1,10 @@
+approvedGitRepositories:
+  - "**"
+
+enableScripts: true
+
+nodeLinker: node-modules
+
+npmMinimalAgeGate: 0
+
+yarnPath: .yarn/releases/yarn-4.18.0.cjs
```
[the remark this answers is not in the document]
tmilost: adressed

frontend/Dockerfile
```diff
@@ -1,7 +1,10 @@
 FROM node:lts-krypton AS  build-stage
 WORKDIR /app
-COPY . .
+COPY package.json yarn.lock .yarnrc.yml ./
```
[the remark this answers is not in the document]
tmilost: node:lts-krypton resolves to Node v24.20.0 (four majors above the >=20 floor, not below it), and we already have a real, verified Docker build proving Yarn 4.18.0 resolves and installs cleanly inside this exact image (Step 9's test: 10s install, Yarn 4.18.0 in the log, successful build).

frontend/Dockerfile
```diff
@@ -1,7 +1,10 @@
...
+COPY .yarn ./.yarn
+ENV CYPRESS_INSTALL_BINARY=0
 RUN yarn install
```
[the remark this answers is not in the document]
tmilost: Docker's --immutable is coming — it's staged as a deliberate follow-up commit once this PR's CI run is fully green (see the migration plan), not applied now on purpose, so a lockfile-drift failure doesn't get conflated with the COPY-layer restructuring. corepack enable isn't needed though: node:lts-krypton already has global classic Yarn 1.22.22 on PATH, and we've verified across multiple real builds that it correctly delegates to the pinned Berry 4.18.0 via yarnPath with no Corepack involved.

frontend/.yarn/releases/yarn-4.18.0.cjs
stetsche: Do we need the entire yarn release committed?

frontend/package.json
```diff
@@ -3,6 +3,9 @@
   "version": "0.1.0",
   "private": true,
   "type": "module",
+  "engines": {
+    "node": ">=20"
```
stetsche: Could we bump to 24 right away? Node 20 is EOL.

.github/workflows/CI.yml
```diff
@@ -50,8 +50,17 @@ jobs:
           cache: 'yarn'
           cache-dependency-path: 'frontend/yarn.lock'
 
+      - name: Guard against the Yarn pin regressing
```
stetsche, line 53: Might benefit from `working-directory: frontend`

frontend/README.md
```diff
@@ -4,6 +4,12 @@ This project uses Yarn as the package manager.
 Yarn is recommended because some legacy packages specify peer
 dependencies that can conflict with newer package versions when using npm.
 
+Yarn is pinned via `.yarnrc.yml` / `.yarn/releases` to version 4.18.0 (Berry).
```
stetsche: I wouldn't put the current version in the doc. Then we either need to update it there all the time too, or it will dirft.

frontend/.dockerignore
```diff
@@ -1,3 +1,6 @@
 .github
 tests
 docs
+node_modules
```
stetsche, line 4: Does `node_modules` still exist?

frontend/.yarnrc.yml
```diff
@@ -0,0 +1,7 @@
+enableScripts: true
+
+nodeLinker: node-modules
```
stetsche, line 3: Isn't the default the main argument for the upgrade? https://yarnpkg.com/features/linkers#nodelinker-pnp

stetsche (changes requested):

frontend/.yarn/releases/yarn-4.18.0.cjs
stetsche, line 1: If I understand right, using [corepack](https://yarnpkg.com/corepack), which comes with node, we wouldn't need to commit the yarn release. Sounds favorable to me. What do you think?

frontend/.dockerignore
tmilost: If you mistakenly run `yarn install` with Yarn 1.22 on your local machine

stetsche: With the PR checked out and yarn 4, it is still created, as it is configured `.yarnrc.yml`.
```
$ yarn --version
4.18.0
$ ls node_modules
ls: cannot access 'node_modules': No such file or directory
markus@DELLSTETSCHNIG:~/repos/bbmri-negotiator/frontend$ yarn install
➤ YN0000: · Yarn 4.18.0
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed
➤ YN0000: ┌ Post-resolution validation
➤ YN0002: │ negotiator-v3-frontend@workspace:. doesn't provide oidc-client (pc35528), requested by pinia-oidc.
➤ YN0086: │ Some peer dependencies are incorrectly met by your project; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code.
➤ YN0086: │ Some peer dependencies are incorrectly met by dependencies; run yarn explain peer-requirements for details.
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0013: │ 180 packages were added to the project (+ 127.1 MiB).
➤ YN0000: └ Completed in 4s 828ms
➤ YN0000: ┌ Link step
➤ YN0007: │ core-js@npm:3.49.0 must be built because it never has been before or the last one failed
➤ YN0007: │ cypress@npm:15.17.0 must be built because it never has been before or the last one failed
➤ YN0007: │ @parcel/watcher@npm:2.5.6 must be built because it never has been before or the last one failed
➤ YN0000: └ Completed in 39s 497ms
➤ YN0000: · Done with warnings in 44s 490ms
$ ls node_modules | head -n 4
abbrev
acorn
acorn-jsx
agent-base
```

frontend/.yarnrc.yml
tmilost: Given the codebase already has known peer-dependency/hoisting friction under node-modules (vue3-form-wizard, vuedraggable, the @codemirror/lang-vue/codemirror/vue-codemirror trio, vue-router), and PnP is strictly less forgiving about phantom dependencies than node-modules, switching linkers here would likely surface a lot more breakage bundled into the same PR as the version pin.

I'd rather track a PnP switch as its own follow-up with dedicated testing, so it's easy to isolate and revert if something doesn't resolve cleanly. OK to keep this PR scoped to the version pin?

frontend/.yarn/releases/yarn-4.18.0.cjs
tmilost: Corepack only kicks in if someone has already run corepack enable in that environment. If they haven't, a plain yarn command just falls back to whatever Yarn is on their PATH — could be the wrong version, could be missing entirely, and it fails silently. That's the exact bug this migration is fixing, so relying on Corepack would bring the risk right back.

Committing the binary means Yarn resolves to the right version everywhere — local machines, CI, Docker — with no setup step needed. It also lets us add a one-line CI check that fails loudly if the binary or config ever go missing.

One more thing: Node itself is moving Corepack to opt-in rather than bundling it by default, so I'd rather not depend on it being there.

We can still keep Corepack documented as a fallback for someone setting up a machine from scratch — that's already the case.

frontend/.dockerignore
tmilost: sry wrong comment, not switching to PnP in this PR

frontend/package.json
```diff
@@ -65,5 +67,6 @@
     "last 2 versions",
     "not dead",
     "not ie 11"
-  ]
+  ],
+  "packageManager": "yarn@4.18.0"
```
stetsche: I'd say let's remove this, to maintain the version only in `.yarnrc.yml`
```suggestion
  ]
```

stetsche (approved):

frontend/package.json
tmilost: done

PR #1299, opened 2026-09-16
feat(webhook): add negotiation.updated webhook event 

ondraholis:
### Description:

Adds a new `negotiation.updated` webhook event that is triggered  when a negotiation is updated via `PATCH /v3/negotiations/{id}`. The webhook payload contains only the `negotiationId`. 

### Checklist:

**If applicable to this PR:**


backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationServiceImpl.java
```diff
@@ -207,6 +207,9 @@ public NegotiationDTO update(String negotiationId, NegotiationUpdateDTO updateDT
       negotiationEntity.setCreatedBy(person);
     }
     negotiationRepository.saveAndFlush(negotiationEntity);
+    if (Objects.nonNull(updateDTO.getPayload())) {
```
stetsche: `Objects.nonNull(updateDTO.getPayload()` is already checked above. I think we should check instead, if there actually was a change in the payload. Maybe the `displayId` changed? Or maybe we should publish `negotiation.updated`? I'm not sure right now, WDYT?

stetsche (changes requested):

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationServiceImpl.java
ondraholis: Good point. I replaced the negotiation.payload.updated webhook by negotiation.updated webhook. The event is published only when some change in negotiation really happened. I also move the save into the repository to the same guard along with publish event. Please check 
[9b68e45](https://github.com/BBMRI-ERIC/negotiator/pull/1299/commits/9b68e45ed41196386e100e951f69825ecf610226)

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationServiceImpl.java
```diff
@@ -228,14 +228,19 @@ private Negotiation buildNegotiation(NegotiationCreateDTO negotiationBody, Reque
   public NegotiationDTO update(String negotiationId, NegotiationUpdateDTO updateDTO) {
     Negotiation negotiationEntity = findEntityById(negotiationId, true);
     verifyWriteAccessToNegotiation(negotiationEntity);
-    if (Objects.nonNull(updateDTO.getPayload())) {
+    boolean isPayloadUpdated = isPayloadUpdated(updateDTO, negotiationEntity);
```
stetsche: I would keep this code mostly what it was and add one boolean `updated` on which only the event publishing depends on. Something like:
```
boolean updated = false;

if (Objects.nonNull(...) {
    set...
    updated = true;
}
```
I see you are trying to improve the quality here, but
1) We risk introducing bugs in the future when we need to check the boolean at two places.
2) `isSomethingUpdated` can be confusing. We are not checking if payload "has changed between the two objects?", we are checking "should we update based on the dto PATCH".
3) We probably don't want to re-use this methods due to their distinct behaviour

backend/src/main/java/eu/bbmri_eric/negotiator/webhook/event/NegotiationUpdatedWebhookEvent.java
```diff
@@ -0,0 +1,16 @@
...
+ * Webhook data payload for an updated negotiation.
+ *
+ * @param negotiationId identifier of the affected negotiation
+ */
+@WebhookEventDoc(
+    summary = "Negotiation updated",
+    description = "Sent when a negotiation was updated.")
+@Schema(description = "Webhook data payload for an updated negotiation.")
+record NegotiationUpdatedWebhookEvent(
+    @Schema(description = "Identifier of the affected negotiation", example = "negotiation-1")
```
stetsche:
```suggestion
    @Schema(description = "Identifier of the affected negotiation", example = "29f1819b-a1d4-48b6-a284-2f9e269bbd05
")
```

stetsche (changes requested):

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/NegotiationServiceImpl.java
ondraholis: Alright, I removed the helper methods in [c08f627](https://github.com/BBMRI-ERIC/negotiator/pull/1299/commits/c08f627b051c13e9319a376aaa3df1e04d003fbd)

backend/src/main/java/eu/bbmri_eric/negotiator/webhook/event/NegotiationUpdatedWebhookEvent.java
ondraholis: Done in [d94810c](https://github.com/BBMRI-ERIC/negotiator/pull/1299/commits/d94810cbfa2d67ea30d5daecc5fc51dd15e7102e)

stetsche (approved):

PR #1300, opened 2026-09-16
feat(webhook): publish negotiation.post.added webhook on lifecycle transition with message 

ondraholis:
### Description:

When a negotiation state transition is triggered with an optional message via `PUT /v3/negotiations/{id}/lifecycle/{event}`, the message was saved as a post directly through `postRepository.save()` in                                                                                                                       `PersistStateChangeListener`, bypassing `PostServiceImpl` and therefore never publishing a `NewPostEvent`. As a result, the `negotiation.post.added` webhook was never fired in this case. 

This change publishes `NewPostEvent` from `PersistStateChangeListener.createPostFromMessage()` after the post is                                                                                                                                                           
 saved, so registered webhooks receive the `negotiation.post.added` delivery consistently regardless of whether the post originated from the posts API or from a lifecycle transition.

### Potential Issue!

I also spot the potential issue during the implementation and test writing, that the PersistStateChangeListener can publish the NegotiationStateChangeEvent even when the negotiation state is not changed (when the negotiation is null). I believe that the event should be raised in the updateNegotiationStatus method or if you would like to keep the logic that the event is raised in the end of the method, the updateNegotiationStatus can return a boolean like "isUpdated" and if the isUpdated would be true, it would call the publishChangeEvent method.

### Checklist:

**If applicable to this PR:**


backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/state_machine/negotiation/PersistStateChangeListener.java
```diff
@@ -56,7 +57,10 @@ public void onPersist(
     Negotiation negotiation = getNegotiation(negotiationId);
     updateNegotiationStatus(state, negotiation);
 
-    if (Objects.nonNull(postSenderId) && Objects.nonNull(postBody) && !postBody.isEmpty()) {
+    if (Objects.nonNull(postSenderId)
+        && Objects.nonNull(postBody)
+        && !postBody.isEmpty()
+        && Objects.nonNull(negotiation)) {
```
stetsche: We should probably throw an exception if negotiation is null (before this if), as you noted as a potential issue. But is there any real case where this could happen currently?

stetsche (changes requested):
I think we should not publish post and state change events in parallel, but rather add postId to the statechange event. Generally, I think one action should publish one event. In this case you do a state change accompanied with an message, which get's added as a post. Originally the state change event did include a `post` field, which I removed, because it was unused and had the post body rather then the id. Now would be a good time to add it back as nullable `postId`.

PR #1301, opened 2026-09-17
feat: migrate pinia-oidc → raw oidc-client-ts

### LINKED ISSUE #1108: [FEATURE] Migrate from pinia-oidc to oidc-client-ts
**Brief description.**
For oidc related functionality we rely on the `pinia-oidc` library, which has served us well for a while. The library has not seen a release for a while now and the officially recommended alternative library `vue3-oidc` does not seem to be broadly adopted neither. While those libraries do ease integration a lot, we would benefit from the stability and broad adoption of a library like `oidc-client-ts` for more stability and control long term.

- Removed pinia-oidc library
- OIDC flows and checks are re-implmented using the lower level `oidc-client-ts` library

tmilost:
Drops `pinia-oidc@1.0.10`, which pulled its own undeduped `oidc-client-ts@2.5.0` alongside the app's direct `oidc-client-ts@^3.5.0`. We now own the `UserManager` directly — smaller, auditable auth surface.

## 4 commits

1. `services/oidcUserManager.js` (new) — raw `UserManager` factory, config field names already match oidc-client-ts's casing except `silentRedirectUri` → `silent_redirect_uri`.
2. `store/oidc.js` (rewrite) + `router/oidcGuard.js` (new) + `main.js` rewiring — setup-store hydrates from storage and exposes a `ready` promise the guard awaits (fixes a hard-refresh race that could trigger a spurious silent-renew iframe); `tokenExpired`/`silentRenewError` now go live for the first time (pinia-oidc forced `automaticSilentRenew: false` internally, making them dead code).
3. `silent-renew-oidc.js` (rewrite) — same iframe entry, raw call.
4. `yarn remove pinia-oidc` — zero remaining references repo-wide.

## Bug fix folded into commit 1

`oidcUser.name`/`email` etc. came back missing — pinia-oidc silently forced `loadUserInfo: true`; raw oidc-client-ts defaults it `false`, and this IdP only exposes profile claims via `/connect/userinfo`, not the ID token. Added `loadUserInfo: true` explicitly.

## Security

PKCE confirmed on (network-verified `code_challenge`/S256), storage/CSRF/multi-tab unchanged, `revokeTokensOnSignout` still `false` (pre-existing, out of scope).

## Testing

Tested end-to-end in `yarn dev` and `docker compose up`: login → protected-route return path, hard refresh (stays logged in, no spurious iframe), logout, silent renewal (verified with a 10s token — rotates correctly; note a lifetime under oidc-client-ts's ~60s renewal window causes continuous renewal, not a bug, just don't tune it that short), and confirmed the guard skips `signinSilent()` cleanly when `SILENT_REDIRECT_URI` is unset (docker-compose). Zero console errors throughout.

## Out of scope

`revokeTokensOnSignout: true`, CI rotation test, CSP hardening, dead `src/oidc/` dir cleanup.

**If applicable to this PR:**


frontend/README.md
```diff
@@ -148,4 +148,4 @@ applications. It also includes support for user session and access token managem
 See the configuration references:
 
 - [vuex-oidc](https://github.com/perarnborg/vuex-oidc/wiki)
```
stetsche:
```suggestion
```

frontend/src/silent-renew-oidc.js
```diff
@@ -1,7 +1,8 @@
-import { piniaOidcProcessSilentSignInCallback } from 'pinia-oidc'
-import oidcSettings from './config/oidc'
+import { createUserManager } from './services/oidcUserManager'
 
-piniaOidcProcessSilentSignInCallback(oidcSettings).catch((err) => {
-  // Keep this minimal because this page only runs in a hidden iframe.
-  console.error('OIDC silent renew callback failed', err)
-})
+createUserManager()
```
stetsche, line 3: I imagine the callback and the rest of the app would need to share the same instance of `UserManager`, or does it not? should wer import `userManager` rather than `createUserManager`?

frontend/src/store/oidc.js
```diff
@@ -1,13 +1,55 @@
...
+  }
+
+  return {
+    ready,
+    oidcIsAuthenticated,
+    oidcUser,
+    access_token,
+    authenticateOidc,
+    signOutOidc,
+    oidcSignInCallback,
```
stetsche, line 53: Or should we use that function to call back in `silent-renew-oidc.js`?

frontend/src/silent-renew-oidc.js
tmilost: No, they don't need to be the same instance — the silent-renew page runs in a hidden iframe as its own separate bundle, so the singleton can't be shared across that boundary anyway. Only the settings need to match, which they already do

frontend/src/store/oidc.js
tmilost: No, different OIDC method (signinSilentCallback vs signinRedirectCallback) and different runtime (silent-renew-oidc.html is a standalone Vite entry with no Vue/Pinia, so the store isn't reachable there anyway

PR #1303, opened 2026-09-18
feat(webhook): add postId to the negotiation.state.updated webhook

ondraholis:
### Description:

When a negotiation state transition is triggered with an optional message via `PUT /v3/negotiations/{id}/lifecycle/{event}`, the message was saved as a post directly through `postRepository.save()` in `PersistStateChangeListener`, bypassing `PostServiceImpl` and therefore never publishing a `NewPostEvent`. As a result, the negotiation.post.added webhook was never fired in this case.

To keep the convention that one action can trigger publishing of one new event, I added the postId into the NegotiationStateChangeEvent as an optional field and it is also now included in the negotiation.state.updated webhook payload.

### Checklist:

**If applicable to this PR:**


backend/src/main/java/eu/bbmri_eric/negotiator/webhook/event/NegotiationStateUpdatedWebhookEvent.java
```diff
@@ -24,4 +25,9 @@ record NegotiationStateUpdatedWebhookEvent(
     @Schema(description = "Target state after transition", example = "SUBMITTED")
         NegotiationState toState,
     @Schema(description = "State-machine event that triggered the transition", example = "SUBMIT")
-        NegotiationEvent event) {}
+        NegotiationEvent event,
+    @Schema(
+            description = "Optional identifier of the post submitted along with the state change",
+            nullable = true,
+            example = "post-1")
```
stetsche:
```suggestion
            example = "bb50ca76-daa0-4a1e-9d74-c0ae90922554")
```
Should reflect a real value. I see it's bad on negotiationId too though.

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/state_machine/negotiation/NegotiationStateChangeEvent.java
```diff
@@ -10,17 +11,29 @@ public class NegotiationStateChangeEvent extends ApplicationEvent {
   private final NegotiationState fromState;
   private final NegotiationState toState;
   private final NegotiationEvent event;
+  @Nullable private final String postId;
 
   public NegotiationStateChangeEvent(
```
stetsche: Is that constructor still in use?

stetsche (changes requested):

backend/src/main/java/eu/bbmri_eric/negotiator/webhook/event/NegotiationStateUpdatedWebhookEvent.java
ondraholis: Done in [88183a8](https://github.com/BBMRI-ERIC/negotiator/pull/1303/commits/88183a8e0d1954483e04cd5caee70d5d5f2d49b1)

backend/src/main/java/eu/bbmri_eric/negotiator/negotiation/state_machine/negotiation/NegotiationStateChangeEvent.java
ondraholis: Only in the tests, so It can be deleted. Done in [926dce9](https://github.com/BBMRI-ERIC/negotiator/pull/1303/commits/926dce9cacaa99c839ca6e5b2d2a5745e6c8def7)

stetsche (approved):

